PR #7, opened 2024-07-02
add ocamlformat github action

hannesm:
expectation for this PR is to fail (since the code isn't ocamlformat-formatted).

//cc @PizieDust

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -29,4 +29,11 @@ jobs:
       - name: Format code
         run: |
           find . -name \*\.ml -maxdepth 1 | xargs opam exec -- ocamlformat --inplace
-          git diff --quiet
+
+      - name: Commit and push changes
+        run: |
+          git config --global user.name "Auto-OCamlformat"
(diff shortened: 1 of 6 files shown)


hannesm:
So, this is my proposal. A GitHub action that potentially adds a commit with the autoformatted code. What do you think?

PizieDust:
> So, this is my proposal. A GitHub action that potentially adds a commit with the autoformatted code. What do you think?

This will be amazing. Is the pipeline costly to run? If so, we can always just try to remember to run `ocamlformat` before pushing, else this is a really awesome idea to have the CI format and add a commit.

hannesm:
It took 3 minutes. I think if you've your editor with ocamlformat, that's fine. But if not, don't worry, and the CI will take care of it.

PizieDust:
> It took 3 minutes. I think if you've your editor with ocamlformat, that's fine. But if not, don't worry, and the CI will take care of it.

Okay great. I do have ocamlformat. I guess I just have to run `ocamlformat --inplace` before pushing

PR #8, opened 2024-07-03
do not merge: test ocamlformat thingy

hannesm:


hannesm:
great, this works -- merged the right thing into main.

PR #10, opened 2024-07-11
Implementing Middlewares

PizieDust:
This PR introduces continues the work on authentication. It introduces Middlewares, and a first middleware for authentication.

cc @hannesm


unikernel.ml
```diff
@@ -490,39 +497,186 @@ struct
...
+                match json with
+                | Error (`Msg s) ->
+                    Logs.warn (fun m -> m "Failed to parse JSON: %s" s);
+                    let res =
+                      "{\"status\": 400, \"message\": \"Bad request body\"}"
+                    in
+                    Lwt.return (reply ~content_type:"application/json" res)
+                | Ok json -> (
+                    let validate_email_re =
+                      Re.Perl.re "[a-zA-Z0-9.$_!]+@[a-zA-Z0-9]+\\.[a-z]{2,3}"
```
hannesm, line 579: let's add a TODO here to validate the mail address (with @dinosaure libraries, and maybe later also to send an email with a unique link). for this PR, the code is good :)

unikernel.ml
```diff
@@ -490,39 +497,186 @@ struct
...
+                      else if not (Re.execp validate_email_re email) then
+                        Error "Invalid email address."
+                      else if String.length password < 8 then
+                        Error "Password must be at least 8 characters long."
+                      else Ok "Validation passed."
+                    in
+                    let email =
+                      json
+                      |> Yojson.Basic.Util.member "email"
+                      |> Yojson.Basic.to_string
```
hannesm, line 594: it is not clear to me what the `to_string` does, we want to ensure that the "email" thingy is a `` `String `` -- so in the end I guess we need a json_utils module (since we've already similar code in user_model and albatross_json...

unikernel.ml
```diff
@@ -490,39 +497,186 @@ struct
...
+                    in
+                    let password =
+                      json
+                      |> Yojson.Basic.Util.member "password"
+                      |> Yojson.Basic.to_string
+                    in
+                    let validate_user = validate_user_input email password in
+                    match validate_user with
```
hannesm, line 602: we don't need a binding `validate_user`, but can directly do a `match validate_user_input ~email ~password with` (and I'd use labels since both values are strings, and we don't want to confuse them). :)

unikernel.ml
```diff
@@ -490,39 +497,186 @@ struct
...
+                                store := store';
+                                let res =
+                                  "{\"status\": 200, \"success\": true, \
+                                   \"message\": {\"user\": "
+                                  ^ Yojson.Basic.to_string
+                                      (User_model.user_to_json user)
+                                  ^ "}}"
+                                in
+                                let cookie =
+                                  List.find
```
hannesm, line 634: with this, we're very sure that such a cookie exists? I'd suggest to use a List.find_opt, and later revise the data structure to always have such a field :)

unikernel.ml
```diff
@@ -490,39 +497,186 @@ struct
...
+                Lwt.return (reply ~content_type:"application/json" "Dashboard"))
+              reqd
+        | "/users" ->
+            let _, (t : Storage.t) = !store in
             Lwt.return
-              (reply ~content_type:"text/html"
-                 "This is going to be the Dashboard :-)")
+              (reply ~content_type:"application/json"
+                 (Yojson.Basic.to_string (Storage.t_to_json t)))
         | "/unikernel/create" ->
```
hannesm, line 680: here we as wel want to use the middleware to authenticate / ensure there's such a user

user_model.ml
```diff
@@ -101,13 +161,84 @@ let generate_token ?(expires_in = 3600) () =
...
+      match c.name = cookie.name with true -> cookie | false -> c)
+    cookies
+
+let login_user ~email ~password users =
+  let user = check_if_user_exists ~email users in
+  match user with
+  | None -> Error (`Msg "This account does not exist.")
+  | Some u -> (
+      let pass = hash_password password u.uuid in
+      match u.password = pass with
```
hannesm, line 234: let's use String.equal here (and not the polymorphic compare)

user_model.ml
```diff
@@ -101,13 +161,84 @@ let generate_token ?(expires_in = 3600) () =
...
+let find_user_by_key (uuid : string) (user_map : (string * user) list) :
+    user option =
+  List.assoc_opt uuid user_map
+
 let create_user ~name ~email ~password =
   let uuid = Uuidm.to_string (generate_uuid ()) in
   let password = hash_password password uuid in
+  let auth_token = generate_token () in
+  let session =
+    generate_cookie ~name:"molly_session" ~expires_in:604800 ~uuid ()
```
hannesm, line 196: sonce this 604800 is used in several places, let's introduce a top-level binding `let week = 604800 (* a week = 7 days * 24 hours * 60 minutes * 60 seconds *)` and use that

middleware.ml
```diff
@@ -0,0 +1,84 @@
...
+          | Some user -> (
+              let user_session =
+                List.find_opt
+                  (fun (cookie : User_model.cookie) ->
+                    cookie.name = "molly_session")
+                  user.cookies
+              in
+              match user_session with
+              | Some cookie -> (
+                  match cookie.value = value with
```
hannesm, line 61: I'd use `String.equal` here

hannesm:
overall looks great (on my mobile phone), do we need to check the expiration of the cookie somewhere? (fine to put this as todo, but I think we need to encode the timestamp when the cookie was issued to compare `now - expiration_duration < issuing_timestamp`.

I have no idea why the ocamlformat CI errored.

unikernel.ml
PizieDust: oh this is just a helper endpoint for me to check when data changes. I must have committed on error :)

hannesm:
thanks, merged. subsequent changes can follow in new prs

PR #11, opened 2024-07-16
Dashboard UI

PizieDust:
This PR adds the dashboard layout. This layout will be a container that houses other pages. For now, it's configured such that we can pass content to it (other html pages) to be rendered within the dashboard....and some minor refactoring.

### DIFF
--- a/utils.ml
+++ b/utils.ml
@@ -14,8 +14,6 @@ module Email = struct
     match Emile.of_string (Json.clean_string email) with
     | Ok _ -> true
     | Error s ->
-        Logs.err (fun m ->
-            m "Emile-Email-Validation %s"
-              (Format.asprintf "%a\n" Emile.pp_error s));
+        Logs.err (fun m -> m "Emile-Email-Validation: %a" Emile.pp_error s);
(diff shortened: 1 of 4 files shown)


utils.ml
```diff
@@ -1,4 +1,21 @@
...
+end
+
 module Email = struct
   let validate_email email =
-    match Emile.of_string email with Ok _ -> true | Error _ -> false
+    match Emile.of_string (Json.clean_string email) with
+    | Ok _ -> true
+    | Error s ->
+        Logs.err (fun m ->
+            m "Emile-Email-Validation %s"
```
hannesm, line 18: you can use "%a" directly here, and then Emile.pp_error -- no need for the asprintf (neither for the \n - which each log message has at the end)

utils.ml
```diff
@@ -1,4 +1,21 @@
...
+              (Format.asprintf "%a\n" Emile.pp_error s));
```
hannesm, line 19: as said above, this can be dropped

hannesm:
merged, thanks

PR #13, opened 2024-07-26
Verify Email middleware

PizieDust:
This PR adds a new middleware for checking if a user's email has been verified. 
There's also addition of smaller utility function such as checking if the cookie in the header is valid and hasn't expired.


middleware.ml
```diff
@@ -1,25 +1,63 @@
...
-let apply_middleware middlewares handler =
-  List.fold_right (fun middleware acc -> middleware acc) middlewares handler
+let apply_middleware ~now middlewares handler =
+  List.fold_right
+    (fun middleware acc -> middleware ~now acc)
+    middlewares handler
 
-let redirect_to_login reqd ?(msg = "") () =
-  let headers = Httpaf.Headers.of_list [ ("location", "/sign-in") ] in
+let get_expired_cookie ~now =
```
hannesm, line 23: I've to admit I don't quite understand why this function is around...

middleware.ml
```diff
@@ -30,55 +68,60 @@ let redirect_to_dashboard reqd ?(msg = "") () =
...
-                      Logs.err (fun m ->
-                          m
-                            "auth-middleware: Session value doesn't match user \
-                             session %s\n"
-                            value);
-                      redirect_to_login reqd ())
-              | None ->
+let cookie_value_from_auth_cookie ~cookie =
+  let parts = String.trim cookie |> String.split_on_char '=' in
+  List.nth parts 1
```
hannesm, line 73: can we have a `match String.split_on_char ..` and some error handling (logging?)

hannesm:
Looks good on general, minir nits above.

From the bird perspective, I lack some understanding why cookie mahic is done in middleware and also in the unikernel. Maybe we need to draw some diagrams on what a http request should have for validation? I'm still a big fan of e.g. webmachine diagram https://raw.githubusercontent.com/wiki/webmachine/webmachine/images/http-headers-status-v3.png

(also there's http://github.com/inhabitedtype/ocaml-webmachine/ but rather dated an stuck with cohttp) -- so let's have a simple version thereof for mollymawk :)

for functions with only a single argument, I'd not use labels. Usually I use them if I pass two arguments of the same type, or if I pass boolean values.

middleware.ml
PizieDust: You are correct, there's no reason to have this function.

PizieDust:
> Looks good on general, minir nits above.
> 
> From the bird perspective, I lack some understanding why cookie mahic is done in middleware and also in the unikernel. Maybe we need to draw some diagrams on what a http request should have for validation? I'm still a big fan of e.g. webmachine diagram https://raw.githubusercontent.com/wiki/webmachine/webmachine/images/http-headers-status-v3.png
> 
> (also there's [http://github.com/inhabitedtype/ocaml-webmachine/](https://github.com/inhabitedtype/ocaml-webmachine/) but rather dated an stuck with cohttp) -- so let's have a simple version thereof for mollymawk :)
> 
> for functions with only a single argument, I'd not use labels. Usually I use them if I pass two arguments of the same type, or if I pass boolean values.

Thank you for the review and this diagram. It is really helpful. I will at some point create a dedicated error page that we can call to handle all the different types of errors, and also be sure to include the appropriate status codes.

user_model.ml
```diff
@@ -268,32 +274,38 @@ let find_user_by_key (uuid : string) (user_map : (string * user) list) :
     user option =
   List.assoc_opt uuid user_map
 
-let create_user ~name ~email ~password ~created_at =
+let create_user name email password created_at =
```
hannesm, line 277: here I'd retain the labels

user_model.ml
```diff
@@ -240,7 +246,7 @@ let generate_token ?(expires_in = 3600) ~created_at () =
     created_at;
   }
 
-let generate_cookie ~name ~uuid ?(expires_in = 3600) ~created_at () =
+let generate_cookie name uuid ?(expires_in = 3600) created_at () =
```
hannesm, line 249: I'd retain the labels here as well

user_model.ml
```diff
@@ -268,32 +274,38 @@ let find_user_by_key (uuid : string) (user_map : (string * user) list) :
...
+let update_user user ?name ?email ?email_verified ?password ?tokens ?cookies ()
+    =
   {
     user with
     name = (match name with Some name -> name | _ -> user.name);
     email = (match email with Some email -> email | _ -> user.email);
+    email_verified =
+      (match email_verified with
+      | Some email_verified -> email_verified
+      | _ -> user.email_verified);
```
hannesm, line 308:
```suggestion
      Option.value ~default:user.email_verified email_verified;
```

user_model.ml
```diff
@@ -306,17 +318,26 @@ let update_cookies (cookies : cookie list) (cookie : cookie) : cookie list =
...
+  Utils.TimeHelper.diff_in_seconds cookie.created_at now < cookie.expires_in
+
+let is_email_verified user = Option.is_some user.email_verified
+
+let user_auth_cookie_from_user (user : user) =
+  List.find_opt
+    (fun (cookie : cookie) -> String.equal cookie.name "molly_session")
+    user.cookies
+
+let login_user email password users now =
```
hannesm, line 331: again, I'd use labels for `~email` and `~password`

utils.ml
```diff
@@ -40,4 +40,18 @@ module TimeHelper = struct
   let print_human_readable ~now ~timestamp =
     let duration = diff_in_seconds now timestamp |> Duration.of_sec in
     Format.asprintf "%a" Duration.pp duration
+
+  (* parse Ptime.t option to json *)
+  let ptime_to_json = function
+    | Some ptime -> `String (string_of_ptime ptime)
+    | None -> `Null
```
hannesm, line 47: use `Option.value`

verify_email.ml
```diff
@@ -0,0 +1,103 @@
...
+                                   "py-3 px-2 bg-primary-500 rounded \
+                                    text-gray-100 font-semibold cursor-pointer";
+                                 ];
+                             ]
+                           [ txt "Resend Verification Email" ];
+                       ];
+                   ];
+               ];
+             Footer_layout.footer;
+             Tyxml.Html.Unsafe.data "";
```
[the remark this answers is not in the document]
PizieDust: This will be corrected with a function to call our backend when the user wants to resend a verification email, in case the one they received already expired. :)

utils.ml
```diff
@@ -55,3 +46,49 @@ module TimeHelper = struct
...
+        username = "2b08693767f1d9";
+        password = "55168d154ddc56";
+      }
+    in
+    let authentication =
+      {
+        Sendmail.mechanism = PLAIN;
+        username = config_data.username;
+        password = config_data.password;
+      }
```
hannesm, line 91: this should be part of the configuration -- either passed via command line arguments or stored in the persistent storage...

utils.ml
```diff
@@ -55,3 +46,49 @@ module TimeHelper = struct
...
+    let timestamp_str = TimeHelper.string_of_ptime timestamp in
+    Base64.encode_string
+      (Cstruct.to_string
+         (Mirage_crypto.Hash.SHA256.digest
+            (Cstruct.of_string (uuid ^ "-" ^ timestamp_str))))
+
+  let generate_verification_link uuid timestamp =
+    let signature = generate_signature uuid timestamp in
+    let encoded_uuid = Base64.encode_string uuid in
+    "/auth/verify/token=" ^ encoded_uuid ^ "/" ^ signature
```
hannesm, line 75: the issue with this is: consider we know the user uuid (passed in here) -- and I'm at the moment not sure whether it is exposed (accidentally or as part of an API), we can do the entire token computation.

this means that a person who want to verify an email address they don't have access to, can do so - without receiving the email.

PR #17, opened 2024-08-13
Configuration (albatross endpoint and certificates) is on the block device now

hannesm:
Still missing: a way to upload (and edit) the configuration

The question is how to proceed with the missing part, since we kind of need it before mollymawk is useful again ;)


hannesm:
This will be delayed a bit, let's wait for #19 to land - which will make this PR easier. I converted this PR to a draft and will rebase once we're at the point.

hannesm:
rebased on main, the missing piece is the web form.

PizieDust:
@hannesm I just pushed some commits for this PR. A new menu has been added to the sidebar on the dashboard `Settings`. This is where you will find the configuration UI's

settings_page.ml
```diff
@@ -0,0 +1,483 @@
+let settings_layout ~(configuration : Configuration.t) =
```
hannesm, line 1: why passing this as labeled argument?

configuration.ml
```diff
@@ -0,0 +1,112 @@
...
+
+let current_version = 1
+
+type t = {
+  certificate : X509.Certificate.t;
+  private_key : X509.Private_key.t;
+  server_ip : Ipaddr.t;
+  server_port : int;
+  version : int;
+  updated_at : Ptime.t;
```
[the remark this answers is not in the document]
PizieDust: Well, since it's a thing that can be updated it's a good idea to keep the update time. So the time here represents the timestamp when an update has been made.

utils.ml
```diff
@@ -10,6 +10,15 @@ module Json = struct
 
   let get key assoc =
     Option.map snd (List.find_opt (fun (k, _) -> String.equal k key) assoc)
+
+  let mask_text text max_length =
```
hannesm, line 14: what is the purpose of `mask_text`? I can see you're using it to display the certificate and private key, but why?

PizieDust: It just adds a bunch of asterisk. It's committed by error.

hannesm:
thanks a lot :)

PR #19, opened 2024-08-13
Albatross work

hannesm:
This takes the `domain` into consideration, and enables the `/unikernel/info/` endpoint with the new interface.

Also removed some configuration items (multiple certificates in cert.pem, server.pem), since in the end we don't actually want them anymore.

This has been tested with our albatross server successfully :D


albatross.ml
```diff
@@ -0,0 +1,299 @@
...
+        continue_reading t name flow
+    | Ok `Eof ->
+        t.console_readers <- Set.remove name t.console_readers;
+        TLS.close flow
+    | Error e ->
+        t.console_readers <- Set.remove name t.console_readers;
+        TLS.close flow >|= fun () ->
+        Logs.err (fun m -> m "received error while reading: %a" TLS.pp_error e)
+
+  let query t ?domain ?(name = ".") cmd =
```
PizieDust, line 232: Some thoughts, 
let's say we have this endpoint `/unikernel/info/domain:name` 
and somehow a user figures out they can edit the link to `/unikernel/info/`, could this function "potentially" return the result which would be the list of all unikernels?

hannesm: We never should have the `domain` exposed in an endpoint. The idea is that the `domain` will always be added by mollymawk depending on the user account. And the client can only choose the unikernel name.

hannesm: At the moment, this is all in flux, since the old code and interface didn't deal with the policies properly. But I'm sure we will adapt the code to always use authentication and then to pass the domain properly to albatross :)

hannesm: What I intended to say: good thinking, this is a very valid question. And indeed we need to be careful to not expose that.

As aside, what we developed in Marrakesh was cutting of the ":" from the unikernel names -- there we didn't bother about domains and the policies. Now, a `/unikernel/info/:robur:roburtls` won't work -- since albatross rejects the `:` in the name (only allows letters, digits, hyphen). In order to access the robur unikernels we need, as done in this PR, the intermediate certificate.

And you are very right, mollymawk has the permissions with the certificate and key it holds to create/destroy _all the unikernels_, so we really need to be careful to not expose functionality meant for other users within the HTTP interface. :)

PizieDust: Okay great. I'm happy to know that the `domain` is added later on at some point during processing and the user can never know what their domain is. I think information like this shouldn't be added to any user endpoints, maybe system admins (`Robur`)

hannesm: Now that we're talking about it, maybe the path to go is to make the `domain` argument mandatory :) I went ahead and pushed 7f7128b

PR #21, opened 2024-08-21
UI updates

PizieDust:
This PR does the following.
- Separates the layout into authenticated and guest layouts.
- Displays error messages in a dedicated UI (different for both authenticated users and guest users.
- A single module to manage responses.
- A page for user management (on the dashboard sidebar menu `Users`)
- Adds cookie authentication to the configurations endpoint. (Login and Register I believe don't require the user to already be authenticated so these are open).


hannesm:
This looks nice. Just for my understanding:
- when `/api/YYY` is requested and an error occurs, a json error is being sent back
- when another page is requested and an error occurs, the error page is displayed

both use the HTTP reply code that is the error code as well!?

PizieDust:
> This looks nice. Just for my understanding:
> 
> * when `/api/YYY` is requested and an error occurs, a json error is being sent back
> * when another page is requested and an error occurs, the error page is displayed
> 
> both use the HTTP reply code that is the error code as well!?

Yes exactly. This is the idea.

PR #23, opened 2024-08-28
Correct typos

PizieDust:
@hannesm The images were a bit too compressed with a significant loss in quality so I exported at about 300px


hannesm:
awesome, thanks!

PR #25, opened 2024-08-31
Refactorings

hannesm:
The main purpose is to remove the levels of indentation in unikernel.ml.

I'm a bit unsure about the authentication, esp on the verify email. My perception is:
- we always want to authenticate the cookie and have a user around
- for JS & images & css it's fine to not being authenticated
- for /sign-up and /sign-in we don't need authentication
- for /api-register also no need for authentication (cookie is pushed)
- for /api/login no authentication (it is special, since only at that point the cookie is pushed)

for verify-email the user needs to be logged in (and it must be the same who's verifying the email address).

does this make sense?


PizieDust:
> The main purpose is to remove the levels of indentation in unikernel.ml.
> 
> I'm a bit unsure about the authentication, esp on the verify email. My perception is:
> 
> * we always want to authenticate the cookie and have a user around
> * for JS & images & css it's fine to not being authenticated
> * for /sign-up and /sign-in we don't need authentication
> * for /api-register also no need for authentication (cookie is pushed)
> * for /api/login no authentication (it is special, since only at that point the cookie is pushed)
> 
> for verify-email the user needs to be logged in (and it must be the same who's verifying the email address).
> 
> does this make sense?

Yes this is perfect.

unikernel.ml
```diff
@@ -856,17 +856,47 @@ struct
...
+    Albatross.query ~domain:user.name albatross ~name
+      (`Console_cmd (`Console_subscribe (`Count 10)))
+    >>= function
+    | Error msg ->
+        Logs.warn (fun m -> m "error querying albatross: %s" msg);
+        let status =
+          {
+            Utils.Status.code = 500;
+            title = "Error";
+            data = "Error while querying Albatross.";
```
PizieDust, line 868: could be helpful to add the error message

```suggestion
data = "Error while querying Albatross:  " ^ msg;
```

PizieDust (changes requested):

PizieDust (approved):

PizieDust:
Thank you @hannesm . I will merge. 🥇

Issue #38, opened 2024-09-05
REST API


PizieDust:
While Mollymawk presents a nice dashboard through which most actions can be performed, it can be very helpful to expose a REST api through which similar actions can be taken, including:

- [x] create and delete authentication tokens from the dashboard
- [ ] authenticate the tokens in requests
- [ ]  unikernel actions (create/update/destroy)

PR #39, opened 2024-09-06
Superadmins

PizieDust:
This PR implements super-users and includes an admin middleware.
To be included is the UI to make other users admins.


unikernel.ml
```diff
@@ -1047,32 +1109,34 @@ struct
                      (Guest_layout.guest_layout
                         ~page_title:"Deploy unikernels with ease | Mollymawk"
                         ~content:Index_page.index_page ~icon:"/images/robur.png"
-                        ())))
+                        ())
+                     `OK))
         | "/main.js" ->
             check_meth `GET (fun () ->
-                Lwt.return (reply reqd ~content_type:"text/plain" js_file))
+                Lwt.return (reply reqd ~content_type:"text/plain" js_file `OK))
```
hannesm, line 1116: while at it, the `content_type` should be `text/javascript` here.

unikernel.ml
```diff
@@ -1093,24 +1157,23 @@ struct
...
-                (* TODO: a middleware for admins *)
-                authenticate !store reqd (toggle_user store reqd))
-        | "/unikernel-info" ->
+                authenticate ~check_admin:true ~api_meth:true !store reqd
+                  (toggle_user store reqd))
+        | "/api/admin/unikernels" ->
             check_meth `GET (fun () ->
-                authenticate !store reqd (unikernel_info !albatross reqd))
+                authenticate ~check_admin:true ~api_meth:true !store reqd
+                  (unikernel_info !albatross reqd))
```
hannesm, line 1176: why is this `unikernel_info` now requiring admin privileges?

users_index.ml
```diff
@@ -143,29 +143,42 @@ let users_index_layout (users : User_model.user list) current_time =
...
-                                                 ~a:
-                                                   [
-                                                     a_class
-                                                       [
-                                                         "text-secondary-500 \
-                                                          fa-solid fa-x";
-                                                       ];
-                                                   ]
-                                                 []);
+                                           (if user.super_user then
```
hannesm, line 146: shouldn't we distinguish, also in the super-user case, between being active and not active?

hannesm:
I pushed a commit to this PR to use mirage < 4.7.0 and albatross < 2.2.0. let's make the upgrade to the next version of mirage & albatross in a separate PR.

unikernel.ml
PizieDust: This endpoint is actually not being used, but I give it the admin privileges as my plan is that it should be an endpoint that list all unikernels from all users.

users_index.ml
PizieDust: Yes I think we can do that.

unikernel.ml
hannesm: I'm not sure, can we defer that to a later time (taking https://github.com/robur-coop/mollymawk/issues/38 into account, which is about the REST API)?

hannesm:
that's fine to merge, or do you plan to include the UI (toggle admin status) in this PR?

unikernel.ml
```diff
@@ -178,7 +187,8 @@ struct
         in
         Lwt.return
           (reply reqd ~content_type:"application/json"
-             (Utils.Status.to_json status))
+             (Utils.Status.to_json status)
+             `Bad_request)
```
hannesm, line 191: I guess esp. here where we have "status" (the json object carrying the reply code) and the polymorphic variant status we can refactor to deduplicate. this can be done in a separate PR.

PizieDust:
> that's fine to merge, or do you plan to include the UI (toggle admin status) in this PR?

I have the base UI ready. Now just need to add the admin toggle functionality. 
It's almost the same code as the activation toggle

PizieDust:
> > that's fine to merge, or do you plan to include the UI (toggle admin status) in this PR?
> 
> I have the base UI ready. Now just need to add the admin toggle functionality. It's almost the same code as the activation toggle

@hannesm I think we can merge now

unikernel.ml
```diff
@@ -316,48 +285,22 @@ struct
...
-                    let status =
-                      {
-                        Utils.Status.code = 200;
-                        title = "Success";
-                        data =
-                          Yojson.Basic.to_string (User_model.user_to_json user);
-                        success = true;
-                      }
-                    in
+
```
hannesm, line 303:
```suggestion
```

unikernel.ml
```diff
@@ -376,36 +318,23 @@ struct
                             ("location", "/dashboard");
                           ]
                         in
-                        Lwt.return
-                          (reply reqd ~header_list
-                             ~content_type:"application/json"
-                             (Utils.Status.to_json status))
+
```
hannesm, line 321:
```suggestion
```

unikernel.ml
```diff
@@ -909,89 +689,44 @@ struct
...
-                            Utils.Status.code = 500;
-                            title = "Error";
-                            data = "Error while querying Albatross.";
-                            success = false;
-                          }
-                        in
-                        reply reqd ~content_type:"application/json"
-                          (Utils.Status.to_json status)
+                            m "Error querying albatross: %s" err);
+
```
hannesm, line 696:
```suggestion
```

unikernel.ml
```diff
@@ -1004,36 +738,65 @@ struct
...
-              success = false;
-            }
-          in
-          Lwt.return
-            (reply reqd ~content_type:"application/json"
-               (Utils.Status.to_json status))
+          http_response reqd ~status_code:400 ~title:"Error"
+            ~data:"Bad HTTP request method." ~success:false `Bad_request
         in
+
```
hannesm, line 799:
```suggestion
```

hannesm:
thanks a lot, merged. sorry for the delay.

PR #44, opened 2024-09-12
Check for duplicate names or emails

### LINKED ISSUE #42: Prevent different users from using the same name
Currently multiple accounts can have the exact same name. We require the email to be unique but not the name. 
This can be problematic given that the `domain` used by albatross is the name of the account.

PizieDust:
fixes #42 

This PR checks that multiple accounts don't have the same names or emails. 
This is made on the assumption that albatross domains are case sensitive. So `foo` and `Foo` shouldn't have any issues.


hannesm:
thanks

PR #45, opened 2024-09-22
Handle basic policies

PizieDust:
This PR starts the journey of adding functionality for managing policies via Mollymawk. 
Had a bit of back and forth with myself on what a good design for the UI should be, which took some days to eventually settle on a functional one. 

The PR at this moment just displays the policies for each user (the edit button doesn't work yet). That could be included in this PR or a follow-up pr. 

@hannesm


unikernel.ml
```diff
@@ -754,12 +754,26 @@ struct
                    reply);
              [])
         >>= fun unikernels ->
+        (Albatross.query albatross ~domain:u.name (`Policy_cmd `Policy_info)
```
hannesm, line 757: I am a bit unsure about the policies and our approach for the source of truth... so we have albatross running somewhere, do we expect (and support) that some other means (e.g. command-line) is used for inserting/adding/removing policies?

Or should mollymawk be run under the assumption that only mollymawk is used to maintain albatross? So we can retrieve and cache the policies at startup and are sure they'll not change over time.

As said, I'm undecided - it is mainly a question about how we expect mollymawk to be used...

PR #47, opened 2024-09-23
use cached policy from albatross

hannesm:


hannesm:
the story is that albatross.ml does a connect to the remote albatross daemon on configuration change (and on startup), where all the policies are fetched. the t.policies field in albatross.ml preserves this information, and unikernel.ml uses that :)

this also means with Albatross.policy we have a single function accessing that information, so if we decide to modify our behaviour, we only have to change it at a single place. :)

user_single.ml
```diff
@@ -239,175 +238,183 @@ let user_single_layout (user : User_model.user) unikernels policies current_time
...
-                    [
-                      a_class
-                        [ "table-auto min-w-full divide-y divide-gray-200" ];
-                    ]
-                  ~thead:
-                    (thead
-                       [
-                         tr
+                (match policy with
+                | None -> h2 [ txt "No policy" ]
```
hannesm, line 242: I'm not sure about this line -- I think if we don't see any policy, we don't need to display any.

PizieDust (approved):
Thank you @hannesm

user_single.ml
PizieDust: I think it's fine. We can complement it with a button to set new policies

PR #56, opened 2024-09-30
provide a nicer interface for policies:

hannesm:
policy now optionally takes a ?domain
(new function) policies returns all sub-policies of the given domain

//cc @PizieDust with this you should be able to do what you intent to do.


PizieDust:
Thank you @hannesm . I'll merge this now

PR #57, opened 2024-09-30
Update policies

PizieDust:
This PR adds some new functionality to update policies.


albatross.ml
```diff
@@ -10,6 +10,16 @@ struct
...
   }
 
+  let empty_policy =
+    Vmm_core.Policy.
+      {
+        vms = 0;
+        cpuids = Vmm_core.IS.of_list [];
+        memory = 0;
+        block = Some 0;
+        bridges = Vmm_core.String_set.of_list [];
```
hannesm, line 20:
```suggestion
        bridges = Vmm_core.String_set.empty;
```

albatross.ml
```diff
@@ -10,6 +10,16 @@ struct
     mutable policies : Vmm_core.Policy.t Vmm_trie.t;
   }
 
+  let empty_policy =
+    Vmm_core.Policy.
+      {
+        vms = 0;
+        cpuids = Vmm_core.IS.of_list [];
```
hannesm, line 17:
```suggestion
        cpuids = Vmm_core.IS.empty;
```

albatross.ml
```diff
@@ -38,6 +48,35 @@ struct
     in
     Ok (Vmm_trie.fold path t.policies (fun name p acc -> (name, p) :: acc) [])
 
+  let policy_resource_used () t =
+    let root_policy =
+      match policy t with
+      | Ok p -> ( match p with Some p -> p | None -> empty_policy)
+      | Error _ -> empty_policy
+    in
+    let policies = match policies t with Ok p -> p | Error _err -> [] in
```
hannesm, line 57: In the error case, let's please print a log message.

albatross.ml
```diff
@@ -38,6 +48,35 @@ struct
...
+    let vms_used, memory_used, storage_used =
+      List.fold_left
+        (fun (total_vms, total_memory, total_block) (name, policy) ->
+          if name <> Vmm_core.Name.root then
```
hannesm, line 61:
```suggestion
          if not Vmm_core.Name.(equal name root) then
```

albatross.ml
```diff
@@ -38,6 +48,35 @@ struct
...
+            ( total_vms + policy.Vmm_core.Policy.vms,
+              total_memory + policy.memory,
+              total_block + match policy.block with Some b -> b | None -> 0 )
```
hannesm, line 64:
```suggestion
              total_block + Option.value ~default:0 policy.block)
```

albatross.ml
```diff
@@ -38,6 +48,35 @@ struct
...
+    in
+    Vmm_core.Policy.
+      {
+        vms = root_policy.vms - vms_used;
+        cpuids = root_policy.cpuids;
+        memory = root_policy.memory - memory_used;
+        block =
+          (match root_policy.block with
+          | Some b -> if b > 0 then Some (b - storage_used) else None
+          | None -> None);
```
hannesm, line 76:
```suggestion
        block = Option.map (fun b -> b - storage_used) root_policy.block;
```

I guess the `> 0` check is sensible, but we don't do it for other values.... I just wish we had an unsigned integer type ;)

albatross.ml
```diff
@@ -38,6 +48,35 @@ struct
...
+        bridges = root_policy.bridges;
+      }
+
   let key_ids exts pub issuer =
     let open X509 in
     let auth = (Some (Public_key.id issuer), General_name.empty, None) in
```
hannesm, line 82: further down in this file, in `gen_cert`, we also use an `empty_policy` -- now with the introduction of that value at the top level, we can replace:

```OCaml
          let policy =
            Option.value
              ~default:
                Vmm_core.(
                  Policy.
                    {
                      vms = 0;
                      cpuids = IS.empty;
                      memory = 0;
                      block = None;
                      bridges = String_set.empty;
                    })
              policy
          in
```

by
```OCaml
          let policy = Option.value ~default:empty_policy policy in
```

header_layout.ml
```diff
@@ -22,4 +22,12 @@ let header ?(page_title = "Mollymawk") ~icon () =
           ~href:"https://unpkg.com/aos@2.3.1/dist/aos.css" ();
         script ~a:[ a_src "https://unpkg.com/aos@2.3.1/dist/aos.js" ] (txt "");
         link ~rel:[ `Icon ] ~href:icon ();
+        script
+          ~a:
+            [
+              a_defer ();
+              a_src
+                "https://cdn.jsdelivr.net/npm/alpinejs@3.x.x/dist/cdn.min.js";
```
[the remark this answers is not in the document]
hannesm: (I'm as well unsure why aos is needed, but I'm fine with your choices)

PizieDust:
@hannesm when trying to add a policy I get this error:
```
Error while querying Albatross: eof from albatross 193.30.40.133:1025 querying [vm: pixie:.] policy add policy: 1 vms 4,12 cpus 50 MB memory 100 MB block, bridges: as250
```

albatross.ml
```diff
@@ -38,6 +48,35 @@ struct
     in
     Ok (Vmm_trie.fold path t.policies (fun name p acc -> (name, p) :: acc) [])
 
+  let policy_resource_used () t =
```
hannesm, line 51:
```suggestion
  let policy_resource_used t =
```

no need for the unit argument

header_layout.ml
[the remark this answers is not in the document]
hannesm: thanks for the explanation, no need to remove aos. it would help me to have a comment in the code a la "aos is animate-on-scroll, used for the home page and make cards bouncy" and "alpine makes multiselect nice"

albatross.ml
```diff
@@ -38,6 +48,32 @@ struct
     in
     Ok (Vmm_trie.fold path t.policies (fun name p acc -> (name, p) :: acc) [])
 
+  let policy_resource_used t =
```
hannesm, line 51: the function name is misleading, it returns the unassigned resources, not the used ones (at least for vms, memory, block).

unikernel.ml
```diff
@@ -779,6 +779,128 @@ struct
...
+                  get "cpuids" xs,
+                  get "bridges" xs,
+                  get "user_uuid" xs )
+            with
+            | ( Some (`Int vms),
+                Some (`Int memory),
+                Some (`Int block),
+                Some (`String cpuids),
+                Some (`String bridges),
+                Some (`String user_uuid) ) -> (
```
hannesm, line 846: This json decoding should be in albatross_json.ml.

unikernel.ml
```diff
@@ -779,6 +779,128 @@ struct
...
+                    let policy_data =
+                      {
+                        Vmm_core.Policy.vms;
+                        memory;
+                        block = Some block;
+                        cpuids =
+                          (if cpuids = "" then Vmm_core.IS.empty
+                           else
+                             Vmm_core.IS.of_list
+                               (List.map int_of_string
```
hannesm, line 861: int_of_string may raise an exception, and this is user input -- we need to handle this properly (use a try .. catch or the int_of_string_opt)

unikernel.ml
```diff
@@ -779,6 +779,128 @@ struct
...
+                in
+                match User_model.find_user_by_key user_uuid users with
+                | Some u -> (
+                    let policy_data =
+                      {
+                        Vmm_core.Policy.vms;
+                        memory;
+                        block = Some block;
+                        cpuids =
+                          (if cpuids = "" then Vmm_core.IS.empty
```
hannesm, line 858: A policy with no cpuid isn't very useful.

unikernel.ml
```diff
@@ -779,6 +779,128 @@ struct
...
+                           else
+                             Vmm_core.IS.of_list
+                               (List.map int_of_string
+                                  (String.split_on_char ',' cpuids)));
+                        bridges =
+                          (if bridges = "" then Vmm_core.String_set.empty
+                           else
+                             Vmm_core.String_set.of_list
+                               (String.split_on_char ',' bridges));
+                      }
```
hannesm, line 868: Could we call to `Vmm_core.Policy.usable` once we have created a policy? And on error return that error. This way we don't need to bug albatross with bad policies ;)

hannesm: Also note that in albatross we may need to check that block is not empty....

hannesm:
> @hannesm when trying to add a policy I get this error:
> 
> ```
> Error while querying Albatross: eof from albatross 193.30.40.133:1025 querying [vm: pixie:.] policy add policy: 1 vms 4,12 cpus 50 MB memory 100 MB block, bridges: as250
> ```

That's kind of expected, since the albatross.ml is not yet ready for it... but so far mollymawk has managed to add some policies:

```sh
root@mirage:~ # albatross-client policy
host [vm: :]: [vm: :]: policy: 4611686018427387903 vms 0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18,19 cpus 130918 MB memory 829848 MB block, bridges: as250, internal, management, service
[vm: pixie:]: policy: 0 vms  cpus 0 MB memory no block
[vm: dust:]: policy: 0 vms  cpus 0 MB memory no block
[vm: Robur:]: policy: 0 vms  cpus 0 MB memory no block
[vm: .:]: policy: 0 vms  cpus 0 MB memory no block
```

I'll have a look once I finished a review here...

hannesm:
> I'll have a look once I finished a review here...

see #58, in unikernel.ml then use `Albatross.set_policy` instead of the query `` `Policy_add ``

The reason behind this is that albatross using TLS is a bit tricky, since it embeds the policies in the certificate chain that is transported across... so the policy_add is implicit via providing a nice intermediate certificate.

unikernel.ml
hannesm: will be caught by Policy.usable

unikernel.ml
hannesm: see other comment about usable, I think in albatross_json it is good.

albatross_json.ml
```diff
@@ -173,6 +173,45 @@ let block_device_of_json js =
...
+          Ok
+            {
+              Vmm_core.Policy.vms;
+              memory;
+              block = Some block;
+              cpuids =
+                (if cpuids = "" then Vmm_core.IS.empty
+                 else
+                   Vmm_core.IS.of_list
+                     (List.map int_of_string (String.split_on_char ',' cpuids)));
```
hannesm, line 201: the `int_of_string` still needs to be guarded, since it may raise an exception.

albatross.ml
```diff
@@ -38,6 +48,35 @@ struct
     in
     Ok (Vmm_trie.fold path t.policies (fun name p acc -> (name, p) :: acc) [])
 
+  let policy_resource_used () t =
+    let root_policy =
+      match policy t with
+      | Ok p -> ( match p with Some p -> p | None -> empty_policy)
+      | Error _ -> empty_policy
```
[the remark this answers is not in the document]
PizieDust: should we fail the request at this point?

albatross_json.ml
```diff
@@ -173,6 +173,50 @@ let block_device_of_json js =
...
+              bridges =
+                (if bridges = "" then Vmm_core.String_set.empty
+                 else
+                   Vmm_core.String_set.of_list
+                     (String.split_on_char ',' bridges));
+            }
+          in
+          match Vmm_core.Policy.usable policy with
+          | Ok () -> Ok policy
+          | Error (`Msg err) -> Error (`Msg err))
```
hannesm, line 212:
```suggestion
          let ( let* ) = Result.bind in
          let* () = Vmm_core.Policy.usable policy in
          Ok policy
```

albatross_json.ml
```diff
@@ -173,6 +173,50 @@ let block_device_of_json js =
...
+          let policy =
+            {
+              Vmm_core.Policy.vms;
+              memory;
+              block = (if block = 0 then None else Some block);
+              cpuids =
+                (if cpuids = "" then Vmm_core.IS.empty
+                 else
+                   Vmm_core.IS.of_list
+                     (List.filter_map int_of_string_opt
```
hannesm, line 201: this is nice, but I wonder whether if "int_of_string_opt" returns None should be reported!?

albatross.ml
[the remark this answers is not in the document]
hannesm: I guess so, as mentioned somewhere else, I'd expect this endpoint to fail if there's no albatross connected that gave us a root policy. Since we don't store the policy persistently in mollymawk, we need a working albatross which stores the policy.

unikernel.ml
```diff
@@ -779,6 +779,78 @@ struct
...
+          `Bad_request
+    | Ok json -> (
+        let user_uuid =
+          Yojson.Basic.(to_string (json |> Util.member "user_uuid"))
+        in
+        let users = User_model.create_user_uuid_map (snd store).Storage.users in
+        match User_model.find_user_by_key user_uuid users with
+        | Some u -> (
+            match Albatross_json.policy_of_json json with
+            | Ok policy -> (
```
hannesm, line 837: Here I'd expect yet another check:
```
let root_policy = Albatross.policy albatross in
match Vmm_core.Policy.is_smaller ~super:root_policy ~sub:policy with
| Error `Msg msg -> log an error and return the error message
| Ok () -> .. as here
```

This way we ensure that the policy is usable and doesn't contradict the root policy (e.g. only cpuids that are available are assigned, ...).

unikernel.ml
```diff
@@ -779,6 +779,78 @@ struct
...
+                Albatross.set_policy albatross ~domain:u.name policy
+                >>= function
+                | Error err ->
+                    http_response reqd ~title:"Error" ~data:err
+                      `Internal_server_error
+                | Ok policies ->
```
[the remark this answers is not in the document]
PizieDust: I updated it to return the policy. Given that we call it from an API endpoint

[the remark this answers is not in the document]
hannesm: ok

albatross_json.ml
PizieDust: I updated to log some information but I'm unsure if this is sufficient. Indeed an empty `cpuid` list is useless

PizieDust:
@hannesm What do you think about a reset button for policies? That clears the existing policies and assigns the user an empty policy. 
What happens if a user already deployed some vms and then their policy is reset. For instance, say I have two deployed vms and then my policy is reset and I have 0 vms in the policy. What happens to the two existing deployed vms? Or does the new policy only affect my account moving forward, meaning I can't deploy anymore and the existing two continue to run?

albatross.ml
```diff
@@ -363,7 +403,7 @@ struct
         (* now we tell albatross about it, using a command for throwing it away *)
         (* note that the 'certs' / 'gen_cert' uses the policies for intermediate certificates *)
         query t ~domain (`Unikernel_cmd `Unikernel_info) >|= function
-        | Ok _ -> Ok ()
+        | Ok _ -> Ok (Vmm_trie.collect name t.policies)
```
hannesm, line 406: I struggle with this: could you explain what the plan is with this return value? And why it is a "collect name t.policies"? I'd expect that maybe the policy passed in here (which gets installed to albatross) will be returned in the success case.

albatross_json.ml
```diff
@@ -173,6 +173,57 @@ let block_device_of_json js =
...
+                     (fun s ->
+                       match int_of_string_opt s with
+                       | Some i -> Some i
+                       | None ->
+                           Logs.err (fun m -> m "Invalid CPU id: %s" s);
+                           None)
+                     (String.split_on_char ',' cpuids)
+                 in
+                 if parsed_cpuids = [] then Vmm_core.IS.empty
+                 else Vmm_core.IS.of_list parsed_cpuids);
```
hannesm, line 209:
```suggestion
                 Vmm_core.IS.of_list parsed_cpuids;
```

albatross_json.ml
```diff
@@ -173,6 +173,57 @@ let block_device_of_json js =
...
+              memory;
+              block = (if block = 0 then None else Some block);
+              cpuids =
+                (let parsed_cpuids =
+                   List.filter_map
+                     (fun s ->
+                       match int_of_string_opt s with
+                       | Some i -> Some i
+                       | None ->
+                           Logs.err (fun m -> m "Invalid CPU id: %s" s);
```
hannesm, line 204:
```suggestion
                           Logs.warn (fun m -> m "Ignoring invalid CPU id: %s" s);
```

albatross_json.ml
```diff
@@ -173,6 +173,57 @@ let block_device_of_json js =
...
+                           None)
+                     (String.split_on_char ',' cpuids)
+                 in
+                 if parsed_cpuids = [] then Vmm_core.IS.empty
+                 else Vmm_core.IS.of_list parsed_cpuids);
+              bridges =
+                (if bridges = "" then Vmm_core.String_set.empty
+                 else
+                   Vmm_core.String_set.of_list
+                     (String.split_on_char ',' bridges));
```
hannesm, line 214:
```suggestion
                   Vmm_core.String_set.of_list
                     (String.split_on_char ',' bridges);
```

unikernel.ml
```diff
@@ -779,6 +779,118 @@ struct
...
+    match User_model.find_user_by_key uuid users with
+    | Some u -> (
+        let user_policy =
+          match Albatross.policy albatross ~domain:u.name with
+          | Ok p -> (
+              match p with Some p -> p | None -> Albatross.empty_policy)
+          | Error _ -> Albatross.empty_policy
+        in
+        match Albatross.policy_resource_avalaible albatross with
+        | Ok root_policy ->
```
hannesm, line 793: I find the naming here (root_policy) misleading... can we use some other term (it is the unallocated resources)

unikernel.ml
```diff
@@ -779,6 +779,118 @@ struct
...
+          User_model.find_user_by_key (Utils.Json.clean_string user_uuid) users
+        with
+        | Some u -> (
+            match Albatross_json.policy_of_json json with
+            | Ok policy -> (
+                match Albatross.policy albatross with
+                | Error err ->
+                    Logs.err (fun m -> m "policy: %s" err);
+                    http_response reqd ~title:"Error" ~data:err
+                      `Internal_server_error
```
hannesm, line 860: Here I'd do:

```OCaml
| Ok None ->
                          Logs.err (fun m ->
                            m "policy: root policy can't be null ");
                        http_response reqd ~title:"Error"
                          ~data:"root policy is null" `Internal_server_error
```

unikernel.ml
```diff
@@ -779,6 +779,118 @@ struct
...
+                | Ok root_policy -> (
```
hannesm, line 861:
```suggestion
                | Ok Some root_policy -> (
```

and then we can remove one layer of indentation

unikernel.ml
```diff
@@ -779,6 +779,118 @@ struct
...
+                    match root_policy with
+                    | None ->
+                        Logs.err (fun m ->
+                            m "policy: root policy can't be null ");
+                        http_response reqd ~title:"Error"
+                          ~data:"root policy is null" `Internal_server_error
+                    | Some root_policy -> (
```
hannesm, line 868:
```suggestion
```

unikernel.ml
```diff
@@ -779,6 +779,118 @@ struct
...
+        let users = User_model.create_user_uuid_map (snd store).Storage.users in
+        match
+          User_model.find_user_by_key (Utils.Json.clean_string user_uuid) users
+        with
+        | Some u -> (
+            match Albatross_json.policy_of_json json with
+            | Ok policy -> (
+                match Albatross.policy albatross with
+                | Error err ->
+                    Logs.err (fun m -> m "policy: %s" err);
```
hannesm, line 858:
```suggestion
                    Logs.err (fun m -> m "couldn't retrieve root policy: %s" err);
```

albatross.ml
PizieDust: I just refactored it.

unikernel.ml
```diff
@@ -779,6 +779,118 @@ struct
...
+                            m "policy: root policy can't be null ");
+                        http_response reqd ~title:"Error"
+                          ~data:"root policy is null" `Internal_server_error
+                    | Some root_policy -> (
+                        match
+                          Vmm_core.Policy.is_smaller ~super:root_policy
+                            ~sub:policy
+                        with
+                        | Error (`Msg err) ->
+                            Logs.err (fun m -> m "policy: %s" err);
```
hannesm, line 874:
```suggestion
                            Logs.err (fun m -> m "policy %a is not smaller than root policy %a: %s" Vmm_core.Policy.pp policy Vmm_core.Policy.pp root_policy err);
```

albatross.ml
PizieDust: After setting the policy, a good return could just be the policy itself. and we can pass that on as an api response

unikernel.ml
```diff
@@ -779,6 +779,118 @@ struct
...
+                            http_response reqd ~title:"Error" ~data:err
+                              `Internal_server_error
+                        | Ok () -> (
+                            Albatross.set_policy albatross ~domain:u.name policy
+                            >>= function
+                            | Error err ->
```
hannesm, line 874:
```OCaml
Logs.err (fun m -> m "error setting policy %a for %s: %s" Vmm_core.Policy.pp policy u.name err);
```

update_policy.ml
```diff
@@ -0,0 +1,487 @@
...
+          ~a:[ a_class [ "font-semibold text-xl" ] ]
+          [ txt ("Set Policy for " ^ user.name) ];
+        p ~a:[ a_id "form-alert"; a_class [ "my-4" ] ] [];
+        p ~a:[ a_id "user_id"; a_class [ "hidden" ] ] [ txt user.uuid ];
+        div
+          ~a:[ a_class [ "my-3" ] ]
+          [
+            label
+              ~a:[ a_class [ "block text-sm font-medium" ] ]
+              [ txt "Allowed VMs" ];
```
hannesm, line 16:
```suggestion
              [ txt "Allowed unikernels" ];
```

update_policy.ml
```diff
@@ -0,0 +1,487 @@
...
+            p
+              [
+                txt
+                  ("total available: "
+                  ^ string_of_int root_policy.Vmm_core.Policy.vms);
```
hannesm, line 21: I think we have the numbers slightly wrong, let's assume the root policy allows 10 vms, and the user robur has 5, the user pixie has 3. (so the root policy passed in here has 2 vms)

now, in this interface when editing the policy of robur, I'd expected to see 7, so (root_policy.vms + user_policy.vms)

same for memory & block below.

hannesm:
> @hannesm What do you think about a reset button for policies? That clears the existing policies for an existing user and assigns the user an empty policy.

Sounds good to me.

> What happens if a user already deployed some vms and then their policy is reset. For instance, say I have two deployed vms and then my policy is reset and I have 0 vms in the policy. What happens to the two existing deployed vms? Or does the new policy only affect my account moving forward, meaning I can't deploy anymore and the existing two continue to run?

The code in question is:

https://github.com/robur-coop/albatross/blob/v2.2.0/src/vmm_vmmd.ml#L265-L277

Which calls:

https://github.com/robur-coop/albatross/blob/v2.2.0/src/vmm_resources.ml#L290-L299

So, it checks that the policy is aligned with the ones above and below (let's not care about below for now).

It also calls check_vms

https://github.com/robur-coop/albatross/blob/v2.2.0/src/vmm_resources.ml#L257-L288

Which basically restricts setting a policy that would violate the currently running unikernels (in terms of amount, memory usage, block usage, bridges).

user_single.ml
```diff
@@ -238,183 +238,210 @@ let user_single_layout (user : User_model.user) unikernels policy current_time =
...
+                                           ~a:
+                                             [
+                                               a_class
+                                                 [
+                                                   "px-6 py-2 text-start \
+                                                    text-xs font-bold \
+                                                    text-primary-600 uppercase";
+                                                 ];
+                                             ]
+                                           [ txt "Allowed VMs" ];
```
hannesm, line 273:
```suggestion
                                           [ txt "Allowed unikernels" ];
```

assets/main.js
```diff
@@ -250,3 +250,75 @@ async function toggleUserActiveStatus(uuid) {
...
+	try {
+		buttonLoading(policyButton, true, "Processing...")
+		const response = await fetch("/api/admin/u/policy/update", {
+			method: 'POST',
+			headers: {
+				"Content-Type": "application/json",
+			},
+			body: JSON.stringify(
+				{
+					"vms": Number(vm_count),
```
hannesm, line 294:
```suggestion
					"unikernels": Number(unikernel_count),
```

assets/main.js
```diff
@@ -250,3 +250,75 @@ async function toggleUserActiveStatus(uuid) {
...
+			}
+		},
+		removeItem(index) {
+			this.selected.splice(index, 1);
+		}
+	};
+}
+
+async function updatePolicy() {
+	const vm_count = document.getElementById("f_allowed_vms").innerText;
```
hannesm, line 277:
```suggestion
	const unikernel_count = document.getElementById("f_allowed_unikernels").innerText;
```

update_policy.ml
```diff
@@ -0,0 +1,487 @@
...
+                          "border py-2 px-3 border-primary-500 \
+                           hover:bg-primary-100 rounded-md";
+                        ];
+                      Unsafe.string_attrib "x-on:click" "count++";
+                    ]
+                  [ i ~a:[ a_class [ "fa-solid fa-plus" ] ] [] ];
+                span
+                  ~a:
+                    [
+                      a_id "f_allowed_vms";
```
hannesm, line 47:
```suggestion
                      a_id "f_allowed_unikernels";
```

albatross_json.ml
```diff
@@ -173,6 +173,54 @@ let block_device_of_json js =
       | _, _, _ -> Error (`Msg "couldn't decode json"))
   | _ -> Error (`Msg "bad json, either string or assoc")
 
+let policy_of_json js =
+  match js with
+  | `Assoc xs -> (
+      match
+        Utils.Json.
+          ( get "vms" xs,
```
hannesm, line 181:
```suggestion
          ( get "unikernels" xs,
```

albatross_json.ml
```diff
@@ -173,6 +173,54 @@ let block_device_of_json js =
...
+            get "memory" xs,
+            get "block" xs,
+            get "cpuids" xs,
+            get "bridges" xs )
+      with
+      | ( Some (`Int vms),
```
hannesm, line 187:
```suggestion
      | ( Some (`Int unikernels),
```

albatross_json.ml
```diff
@@ -173,6 +173,54 @@ let block_device_of_json js =
...
+          Some (`Int memory),
+          Some (`Int block),
+          Some (`String cpuids),
+          Some (`String bridges) ) ->
+          let policy =
+            {
+              Vmm_core.Policy.vms;
```
hannesm, line 194:
```suggestion
              Vmm_core.Policy.vms = unikernels;
```

update_policy.ml
PizieDust: Yes, this is the correct way to go. 
right now we only display the difference between root_policy and all other policies.
so robur has 5
pixie has 3
root has 2 left. 
if we want to add say 1 vm to pixie, 
the total available should be 7 and not 2 (2 + robur's 5)?

unikernel.ml
```diff
@@ -779,6 +779,125 @@ struct
...
+                                (Yojson.Basic.to_string
+                                   (Albatross_json.policy_info policy))
+                              `OK))
+                | Ok None ->
+                    Logs.err (fun m -> m "policy: root policy can't be null ");
+                    http_response reqd ~title:"Error"
+                      ~data:"root policy is null" `Internal_server_error
+                | Error err ->
+                    Logs.err (fun m ->
+                        m "policy: an error occured while fetching root policy ");
```
hannesm, line 892:
```suggestion
                        m "policy: an error occured while fetching root policy: %s" err);
```

update_policy.ml
hannesm: IMHO we have unallocated (free) things. And then we have the current user for which we adjust the policy.

So, with the example above, in the mask I'd expect:
- for robur: total available: 7, currently assigned: 5
- for pixie: total available: 5, currently assigned: 2

PizieDust: @hannesm I just refactored the implementation a bit. Does it address this point?

update_policy.ml
```diff
@@ -0,0 +1,517 @@
...
+            label
+              ~a:[ a_class [ "block text-sm font-medium" ] ]
+              [ txt "Allowed unikernels" ];
+            p
+              ~a:[ a_class [ "space-x-5" ] ]
+              [
+                txt
+                  ("total available: "
+                  ^ string_of_int
+                      Vmm_core.Policy.(root_policy.vms - user_policy.vms));
```
hannesm, line 24:
```suggestion
                      Vmm_core.Policy.(root_policy.vms + user_policy.vms));
```

in our discussion, root_policy.vms would be 3, user_policy.vms for robur 5, and we want to display 8 -- not "-2".

update_policy.ml
```diff
@@ -0,0 +1,517 @@
...
+            label
+              ~a:[ a_class [ "block text-sm font-medium" ] ]
+              [ txt "Allowed Memory" ];
+            p
+              ~a:[ a_class [ "space-x-5" ] ]
+              [
+                txt
+                  ("total available: "
+                  ^ string_of_int
+                      Vmm_core.Policy.(root_policy.memory - user_policy.memory)
```
hannesm, line 98:
```suggestion
                      Vmm_core.Policy.(root_policy.memory + user_policy.memory)
```

update_policy.ml
```diff
@@ -0,0 +1,517 @@
...
+            p
+              ~a:[ a_class [ "space-x-5" ] ]
+              [
+                txt
+                  ("total available: "
+                  ^ string_of_int
+                      Vmm_core.Policy.(
+                        match (root_policy.block, user_policy.block) with
+                        | Some root_block, Some user_block ->
+                            root_block - user_block
```
hannesm, line 178:
```suggestion
                            root_block + user_block
```

update_policy.ml
```diff
@@ -0,0 +1,517 @@
...
+                        | Some root_block, None -> root_block
```
hannesm, line 179:
```suggestion
                        match (unallocated_resources.block, user_policy.block) with
                        | Some unallocated, Some user_block ->
                            unallocated + user_block
                        | Some unallocated, None -> unallocated
```

update_policy.ml
```diff
@@ -0,0 +1,517 @@
...
+                  ("total available: "
+                  ^ string_of_int
+                      Vmm_core.Policy.(root_policy.vms - user_policy.vms));
+              ];
+            p
+              [
+                txt
+                  ("total unallocated: "
+                  ^ string_of_int unallocated_resources.Vmm_core.Policy.vms);
+              ];
```
hannesm, line 31: I don't understand why this should be displayed...

If I understand it, your change leads to 3 numbers being displayed:
- for robur: total available: 5, total unallocated: 3, form: 5
- for pixie: total available: 8, total unallocated: 3, form: 2

while I mean that the important numbers are: robur has currently 5 assigned, and can go up to 8 -- pixie has 2 assigned and can go up to 5.

PizieDust:
> hmm, no, I don't understand why you now return the unallocated_resources and root_policy -- and fail to understand what the update_policy page should display.

My understanding is that:
- `root_policy` returns all the resources (unallocated + allocated).
- `unallocated_resources`: this is everything that isn't assigned to any policy
- `user_policy`: this is the policy of any specific user

So let's say we have two users `robur` and `pixie`.
- `root_policy` has 10 vms. 
- robur has 5 vms
- pixie has 3 vms
If robur wants to update the policy of pixie, what will be on the UI is
- `unallocated_resources` : 2 vms. 
- `total_available` : 7 vms

I'm not sure if this approach was what you had in mind?

update_policy.ml
PizieDust: Okay I see. So on the page we should preferably have something like:
max vms: 8
form: 5

and for pixie
max vms: 5
form 3

hannesm:
> So let's say we have two users `robur` and `pixie`.
> 
>     * `root_policy` has 10 vms.
> 
>     * robur has 5 vms
> 
>     * pixie has 3 vms
>       If robur wants to update the policy of pixie, what will be on the UI is
> 
>     * `unallocated_resources` : 2 vms.
> 
>     * `total_available` : 7 vms
> 
> 
> I'm not sure if this approach was what you had in mind?

For me, the interesting numbers would be "I can specify to allocate up to 7" and "I currently have 5".

PizieDust:
> > So let's say we have two users `robur` and `pixie`.
> > ```
> > * `root_policy` has 10 vms.
> > 
> > * robur has 5 vms
> > 
> > * pixie has 3 vms
> >   If robur wants to update the policy of pixie, what will be on the UI is
> > 
> > * `unallocated_resources` : 2 vms.
> > 
> > * `total_available` : 7 vms
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I'm not sure if this approach was what you had in mind?
> 
> For me, the interesting numbers would be "I can specify to allocate up to 7" and "I currently have 5".

great, thank you. I think I understand what you meant earlier.

hannesm:
@PizieDust cool, great. I guess the root_policy can be removed from the update_policy module...

hannesm:
Thanks a lot for your amazing work on this, @PizieDust. :)

PR #58, opened 2024-10-01
add a set_policy function to albatross

hannesm:
/cc @PizieDust we need this to set the policies, since we need to adjust our local cache, and then issue any command with the new policy


PizieDust:
@hannesm Thank you. merging now.

PizieDust (approved):

PR #63, opened 2024-10-11
Middleware: respond with content-length

### LINKED ISSUE #62: /sign-in hangs when already logged in
Using the web developer tools I copied the request "as curl" and added `-v`. The response part is below:

```
> 
< HTTP/1.1 302 Found
< location: /dashboard
* no chunk, no close, no size. Assume close to signal end
< 
```

It seems `Httpaf.Reqd.respond_with_string` will look at the headers to determine if it should close or persist the connection: https://ocaml.org/p/httpaf/latest/doc/Httpaf/Reqd/index.html

I'm not sure if this is a bug in httpaf, but I think we can work around this by setting ~~`content-size`~~ `content-length`` in the response.

reynir:
Otherwise, httpaf lets the connection hang and the other end waits for TCP close until timeout /o\ Fixes #62 (I hope)

### DIFF
--- a/middleware.ml
+++ b/middleware.ml
@@ -25,6 +25,7 @@ let redirect_to_login reqd ?(msg = "") () =
         "molly_session=;Path=/;HttpOnly=true;Expires=2023-10-27T11:00:00.778Z"
       );
       ("location", "/sign-in");
+      ("Content-Length", string_of_int (String.length msg));
     ]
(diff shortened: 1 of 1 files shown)


PizieDust (approved):

PizieDust:
Thank you @reynir

PR #65, opened 2024-10-11
Csrf tokens

### LINKED ISSUE #40: CSRF tokens
we need them for all the forms. emit them and check them on receiving form data.

PizieDust:
closes #40 
This PR adds functionality for CSRF token.

When a user is authenticated, we use a [Double Synchronizer Pattern](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#synchronizer-token-pattern) which keeps the token on the server and transmit a copy as hidden field to the form. 

When a user is not authenticated, we use a [Naive Double Submit Cookie Pattern](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#naive-double-submit-cookie-pattern-discouraged) which sends the token in a cookie and also a hidden field on the form.

We also add a custom header `X-MOLLY-CSRF`. Will experiment with this to see if it's better for unauthenticated users than cookies.

### DIFF
--- a/middleware.ml
+++ b/middleware.ml
@@ -187,14 +187,12 @@ let is_user_admin_middleware api_meth now users handler reqd =
           `Unauthorized user 401 api_meth reqd ()
   | Error (`Msg msg) -> redirect_to_login ~msg reqd ()
 
-let csrf_match ~input_csrf ~check_csrf = String.equal input_csrf check_csrf
+let csrf_match ~input_csrf ~check_csrf =
+  String.equal (Utils.Json.clean_string input_csrf) check_csrf
 
 let csrf_cookie_verification form_csrf reqd =
   match has_cookie "molly_csrf" reqd with
-  | Some cookie ->
-      csrf_match
-        ~input_csrf:(Utils.Json.clean_string form_csrf)
-        ~check_csrf:cookie
+  | Some cookie -> csrf_match ~input_csrf:form_csrf ~check_csrf:cookie
   | None -> false
 
 let csrf_verification users now form_csrf handler reqd =
@@ -210,8 +208,7 @@ let csrf_verification users now form_csrf handler reqd =
       | Some csrf_token ->
           if
             User_model.is_valid_cookie csrf_token now
-            && csrf_match ~check_csrf:csrf_token.value
-                 ~input_csrf:(Utils.Json.clean_string form_csrf)
+            && csrf_match ~check_csrf:csrf_token.value ~input_csrf:form_csrf
           then handler reqd
           else
             http_response ~title:"CSRF Token Mismatch"
--- a/middleware.ml
+++ b/middleware.ml
@@ -1,7 +1,13 @@
 type handler = Httpaf.Reqd.t -> unit Lwt.t
 type middleware = handler -> handler
 
-let has_session_cookie (reqd : Httpaf.Reqd.t) =
+let get_csrf now =
+  User_model.(
+    generate_cookie ~name:"molly_csrf"
+      ~uuid:(Uuidm.to_string (generate_uuid ()))
+      ~created_at:now ~expires_in:3600)
+
+let has_cookie cookie_name (reqd : Httpaf.Reqd.t) =
   let headers = (Httpaf.Reqd.request reqd).headers in
   match Httpaf.Headers.get headers "Cookie" with
   | Some cookies ->
@@ -10,7 +16,7 @@ let has_session_cookie (reqd : Httpaf.Reqd.t) =
         (fun cookie ->
           let parts = String.trim cookie |> String.split_on_char '=' in
           match parts with
-          | [ name; _ ] -> String.equal name "molly_session"
+          | [ name; _ ] -> String.equal name cookie_name
           | _ -> false)
         cookie_list
   | _ -> None
@@ -92,6 +98,14 @@ let redirect_to_dashboard reqd ?(msg = "") () =
   Httpaf.Reqd.respond_with_string reqd response msg;
   Lwt.return_unit
 
+let redirect_to_same_page reqd ?(msg = "") () =
+  let request = Httpaf.Reqd.request reqd in
+  let current_path = request.target in
+  let headers = Httpaf.Headers.of_list [ ("location", current_path) ] in
+  let response = Httpaf.Response.create ~headers `Found in
+  Httpaf.Reqd.respond_with_string reqd response msg;
+  Lwt.return_unit
+
 let cookie_value_from_auth_cookie cookie =
   match String.split_on_char '=' (String.trim cookie) with
   | _ :: s :: _ -> Ok (String.trim s)
@@ -108,7 +122,7 @@ let user_from_auth_cookie cookie users =
       Error (`Msg s)
 
 let user_of_cookie users now reqd =
-  match has_session_cookie reqd with
+  match has_cookie "molly_session" reqd with
   | Some auth_cookie -> (
       match user_from_auth_cookie auth_cookie users with
       | Ok user -> (
@@ -161,3 +175,39 @@ let is_user_admin_middleware api_meth now users handler reqd =
             "You don't have the necessary permissions to access this service."
           `Unauthorized user 401 api_meth reqd ()
   | Error (`Msg msg) -> redirect_to_login ~msg reqd ()
+
+let csrf_match ~input_csrf ~check_csrf = String.equal input_csrf check_csrf
+
+let csrf_cookie_verification form_csrf reqd =
+  match has_cookie "molly_csrf" reqd with
+  | Some cookie ->
+      csrf_match
+        ~input_csrf:(Utils.Json.clean_string form_csrf)
+        ~check_csrf:cookie
+  | None -> false
+
+let csrf_form_verification users now form_csrf handler reqd =
+  match user_of_cookie users now reqd with
+  | Ok user -> (
+      let user_csrf_token =
+        List.find_opt
+          (fun (cookie : User_model.cookie) ->
+            String.equal cookie.name "molly_csrf")
+          user.User_model.cookies
+      in
+      match user_csrf_token with
+      | Some csrf_token ->
+          if
+            User_model.is_valid_cookie csrf_token now
+            && csrf_match ~check_csrf:csrf_token.value
+                 ~input_csrf:(Utils.Json.clean_string form_csrf)
+          then handler reqd
+          else
+            redirect_to_same_page reqd
+              ~msg:"CSRF token mismatch error. Please referesh and try again."
+              ()
+      | None ->
+          redirect_to_same_page
+            ~msg:"CSRF token mismatch error. Please referesh and try again."
+            reqd ())
+  | Error (`Msg err) -> redirect_to_login ~msg:err reqd ()
--- a/middleware.ml
+++ b/middleware.ml
@@ -223,6 +223,6 @@ let csrf_verification users now form_csrf handler reqd =
               reqd `Bad_request
       | None ->
           http_response
-            ~data:"CSRF token mismatch error. Please referesh and try again."
-            ~title:"CSRF Token Mismatch" reqd `Bad_request)
+            ~data:"Missing CSRF token. Please referesh and try again."
+            ~title:"Missing CSRF Token" reqd `Bad_request)
   | Error (`Msg err) -> redirect_to_login ~msg:err reqd ()
--- a/middleware.ml
+++ b/middleware.ml
@@ -199,7 +199,9 @@ let csrf_cookie_verification form_csrf reqd =
       | Error (`Msg err) ->
           Logs.err (fun m -> m "Error with csrf cookie %s" err);
           false)
-  | None -> false
+  | None -> 
+    Logs.err (fun m -> m "Couldn't find csrf cookie.");
+    false
 
 let csrf_verification users now form_csrf handler reqd =
   match user_of_cookie users now reqd with
--- a/middleware.ml
+++ b/middleware.ml
@@ -98,12 +98,26 @@ let redirect_to_dashboard reqd ?(msg = "") () =
   Httpaf.Reqd.respond_with_string reqd response msg;
   Lwt.return_unit
 
-let redirect_to_same_page reqd ?(msg = "") () =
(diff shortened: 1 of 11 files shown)


middleware.ml
```diff
@@ -1,16 +1,23 @@
...
+
+let has_cookie cookie_name (reqd : Httpaf.Reqd.t) =
   let headers = (Httpaf.Reqd.request reqd).headers in
   match Httpaf.Headers.get headers "Cookie" with
   | Some cookies ->
       let cookie_list = String.split_on_char ';' cookies in
       List.find_opt
         (fun cookie ->
           let parts = String.trim cookie |> String.split_on_char '=' in
+
```
hannesm, line 18:
```suggestion
```

middleware.ml
```diff
@@ -147,3 +173,41 @@ let is_user_admin_middleware api_meth now users handler reqd =
             "You don't have the necessary permissions to access this service."
           `Unauthorized user 401 api_meth reqd ()
   | Error (`Msg msg) -> redirect_to_login ~msg reqd ()
+
+let csrf_match ~input_csrf ~check_csrf =
+  String.equal (Utils.Json.clean_string input_csrf) check_csrf
```
hannesm, line 178: Why is "clean_string" applied here? I still don't quite understand the purpose of `Utils.Json.clean_string`.

middleware.ml
```diff
@@ -78,13 +85,32 @@ let redirect_to_dashboard reqd ?(msg = "") () =
...
+           ])
+        header_list)
+  in
+  let response = Httpaf.Response.create ~headers http_status in
+  Httpaf.Reqd.respond_with_string reqd response data;
+  Lwt.return_unit
+
+let cookie_value_from_cookie cookie =
   match String.split_on_char '=' (String.trim cookie) with
   | _ :: s :: _ -> Ok (String.trim s)
```
hannesm, line 109: Why are there all these `String.trim` around? Does cookie values have lots of whitespaces that have to be ignored?

middleware.ml
```diff
@@ -147,3 +173,41 @@ let is_user_admin_middleware api_meth now users handler reqd =
...
+let csrf_match ~input_csrf ~check_csrf =
+  String.equal (Utils.Json.clean_string input_csrf) check_csrf
+
+let csrf_cookie_verification form_csrf reqd =
+  match has_cookie "molly_csrf" reqd with
+  | Some cookie -> (
+      match cookie_value_from_cookie cookie with
+      | Ok token -> csrf_match ~input_csrf:form_csrf ~check_csrf:token
+      | Error (`Msg err) ->
+          Logs.err (fun m -> m "Error with csrf cookie %s" err);
```
hannesm, line 186:
```suggestion
          Logs.err (fun m -> m "Error retrieving csrf value from cookie %s" err);
```

middleware.ml
```diff
@@ -147,3 +173,41 @@ let is_user_admin_middleware api_meth now users handler reqd =
...
+          false)
+  | None -> false
```
hannesm, line 188:
```suggestion
  | None -> 
    Logs.err (fun m -> m "Couldn't find csrf cookie.");
    false
```

middleware.ml
```diff
@@ -147,3 +173,41 @@ let is_user_admin_middleware api_meth now users handler reqd =
...
+            && csrf_match ~check_csrf:csrf_token.value ~input_csrf:form_csrf
+          then handler reqd
+          else
+            http_response ~title:"CSRF Token Mismatch"
+              ~data:"CSRF token mismatch error. Please referesh and try again."
+              reqd `Bad_request
+      | None ->
+          http_response
+            ~data:"CSRF token mismatch error. Please referesh and try again."
+            ~title:"CSRF Token Mismatch" reqd `Bad_request)
```
hannesm, line 212:
```suggestion
            ~data:"Missing CSRF token. Please referesh and try again."
            ~title:"Missing CSRF Token" reqd `Bad_request)
```

unikernel.ml
```diff
@@ -49,6 +49,41 @@ struct
   module Store = Storage.Make (BLOCK)
   module Map = Map.Make (String)
 
+  let generate_csrf_token ?(update_user = false) ?(user = None) store now =
+    if update_user then
```
hannesm, line 53: Could you please describe here in a comment what "update_user" is used for?

middleware.ml
PizieDust: For some reason when I get a specific json field it is quoted and String.equal would give a failed match

unikernel.ml
```diff
@@ -49,6 +49,41 @@ struct
...
+          let error =
+            {
+              Utils.Status.code = 500;
+              title = "CSRF Token Error";
+              success = false;
+              data = "Error generating csrf. No user";
+            }
+          in
+          Lwt.return (Error error)
+    else Lwt.return (Ok (Middleware.get_csrf now ()).value)
```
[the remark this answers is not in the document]
PizieDust: We could have a helper function which, at the point when we are adding a new cookie, it runs through and gets rid of all expired cookies?

unikernel.ml
```diff
@@ -66,6 +101,24 @@ struct
       ~on_eof:(on_eof f_init);
     finished >>= fun data -> data
 
+  let extract_csrf_token reqd =
+    decode_request_body reqd >>= fun data ->
+    let json =
+      try Ok (Yojson.Basic.from_string data)
+      with Yojson.Json_error s -> Error (`Msg s)
+    in
+    match json with
```
hannesm, line 110:
```suggestion
    match
      try Ok (Yojson.Basic.from_string data)
      with Yojson.Json_error s -> Error (`Msg s)
    with
```

You tend to introduce a lot of let-bindings, and I personally find fewer bindings make the code easier to understand, so I'd go and move the stuff into the `match .. with`. What do you think? (Note: this does not need to be applied here, we can as well go and refactor the codebase)

hannesm: oh, here actually in the bottom you return `json`, so don't mind the above.

unikernel.ml
```diff
@@ -66,6 +101,24 @@ struct
...
+    | Error (`Msg err) ->
+        Logs.warn (fun m -> m "Failed to parse JSON: %s" err);
+        Lwt.return (String.empty, `Null)
```
hannesm, line 113: I don't quite understand why using `String.empty` here? Wouldn't a `Error ()` or a `None` be more in the OCaml spirit?

unikernel.ml
```diff
@@ -92,7 +145,7 @@ struct
     go (Map.empty, []) m
 
   let authenticate ?(email_verified = true) ?(check_admin = false)
-      ?(api_meth = false) store reqd f =
+      ?(api_meth = false) ?(check_csrf = false) ?(form_csrf = "") store reqd f =
```
hannesm, line 147: is check_csrf true only if form_csrf is provided?
is form_csrf provided only if check_csrf is true?

if that is the case, let's use form_csrf and remove check_csrf, and do the checking if form_csrf is provided.

unikernel.ml
```diff
@@ -127,35 +182,33 @@ struct
...
-      (reply reqd ~content_type:"application/json" ~header_list
-         (Utils.Status.to_json status)
-         http_status)
-
   let sign_up reqd =
-    match Middleware.has_session_cookie reqd with
+    let now = Ptime.v (P.now_d_ps ()) in
+    let csrf = Middleware.get_csrf now () in
+    let csrf_cookie = csrf.name ^ "=" ^ csrf.value ^ ";Path=/;HttpOnly=true" in
+    match Middleware.has_cookie "molly_session" reqd with
```
hannesm, line 188: to avoid having to hardcode `"molly_session"` all over, maybe provide a `Middleware.has_session_cookie reqd`?

unikernel.ml
```diff
@@ -127,35 +182,33 @@ struct
...
     | Some cookie -> (
-        match Middleware.cookie_value_from_auth_cookie cookie with
+        match Middleware.cookie_value_from_cookie cookie with
```
hannesm, line 190: I'd rename this to `Middleware.cookie_value cookie`.

unikernel.ml
```diff
@@ -127,35 +182,33 @@ struct
...
         | Ok "" ->
             Lwt.return
               (reply reqd ~content_type:"text/html"
-                 (Sign_up.register_page ~icon:"/images/robur.png" ())
+                 (Sign_up.register_page csrf.value ~icon:"/images/robur.png" ())
```
hannesm, line 194: Should `csrf.value` be labeled / typed as well? `register_page ~csrf:csrf.value`?

unikernel.ml
```diff
@@ -200,60 +255,75 @@ struct
         let password =
           json |> Yojson.Basic.Util.member "password" |> Yojson.Basic.to_string
         in
-        match validate_user_input ~name ~email ~password with
+        let form_csrf =
+          json |> Yojson.Basic.Util.member "form_csrf" |> Yojson.Basic.to_string
```
hannesm, line 258: The `member` may raise, we may want to protect that with a try .. catch.

But maybe we want another time to guard in a separate module all the Yojson usage, to me it looks like the yojson API is a bit hard to use if you don't like to catch exceptions.

unikernel.ml
```diff
@@ -316,42 +385,47 @@ struct
...
-                      `Internal_server_error)))
+                    Middleware.http_response reqd ~title:"Error"
+                      ~data:(String.escaped err) `Internal_server_error)))
 
   let verify_email store reqd user =
-    let email_verification_uuid = User_model.generate_uuid () in
-    let updated_user =
-      User_model.update_user user
-        ~updated_at:(Ptime.v (P.now_d_ps ()))
-        ~email_verification_uuid:(Some email_verification_uuid) ()
```
hannesm, line 339: this "email_verification_uuid", is this still being updated somewhere?

unikernel.ml
```diff
@@ -492,80 +566,96 @@ struct
...
-  let update_settings stack store albatross reqd _user =
-    decode_request_body reqd >>= fun data ->
-    let json =
-      try Ok (Yojson.Basic.from_string data)
-      with Yojson.Json_error s -> Error (`Msg s)
-    in
-    match json with
+  let update_settings json stack store albatross reqd _user =
+    match Configuration.of_json_from_http json (Ptime.v (P.now_d_ps ())) with
+    | Ok configuration_settings -> (
```
hannesm, line 595: this looks very nice :) a beautiful refactorign

unikernel.ml
hannesm: And `?(user = None)` makes this a "user option option" - is this intended? I.e. as you write below, you need to pass `~user:(Some user)` now... if you would make it `?user` here, you could pass `~user:user` (or even `~user`) below.

unikernel.ml
```diff
@@ -1017,19 +1176,15 @@ struct
...
-          when String.(
-                 length path >= 19 && sub path 0 19 = "/unikernel/destroy/") ->
-            check_meth `GET (fun () ->
-                let unikernel_name =
-                  String.sub path 19 (String.length path - 19)
-                in
-                authenticate !store reqd
-                  (unikernel_destroy !albatross unikernel_name reqd))
+                authenticate !store reqd (deploy_form store reqd))
+        | "/unikernel/destory" ->
```
hannesm, line 1183:
```suggestion
        | "/unikernel/destroy" ->
```

hannesm:
Lots of nice additions and refactorings.

I'm still wondering (and that may be because your initial PR comment needs to be updated):
- when is the HTTP header used?
- when is the cookie used?

I guess we need to tackle some other issues (n upcoming refactorings):
- labels and when to use them for function calls
- local bindings vs `match ... with`
- guard yojson functions in a separate module (for safety reasons, whenever we handle user input, expect the exceptional cases)
- String escaping / cleaning / trimming -- also a potential security issue
- where to store data -- esp. should these csrf tokens really be stored on disk?
- this raises the other question about "which data structure to use" (for e.g. the users)

middleware.ml
hannesm: Hmm, ok. Maybe we should open an issue to re-evaluate the `clean_string` and String.trim usages?

middleware.ml
hannesm: (see above, maybe best to leave this for a future)

unikernel.ml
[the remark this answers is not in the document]
hannesm: Yes, maybe related to whether we really wan to store the CSRF cookies on disk or not. And also what data structure(s) we want. Should there be only ever a single CSRF cookie for one user, or could there be multiple CSRF cookies?

unikernel.ml
hannesm: Let's leave this for a separate PR.

hannesm: see #66

unikernel.ml
hannesm: see also #69

unikernel.ml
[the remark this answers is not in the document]
hannesm: let's do this another time, #68

hannesm:
(rebased and force-pushed on main)

unikernel.ml
hannesm: done in bd73bc7 -- it was the case that a `user` was always provided.

hannesm: ehm, [a142cc0](https://github.com/robur-coop/mollymawk/pull/65/commits/a142cc076e4ea7ed057539c991b85a88ba60931e) is the fixed up commit that actually compiles

unikernel.ml
hannesm: done in b4aee8c

unikernel.ml
hannesm: done in 88895ec

unikernel.ml
hannesm: done in 88895ec

unikernel.ml
hannesm: done in eedc956

unikernel.ml
hannesm: done in 883b2bd

hannesm:
IMHO ready to go now

PizieDust:
Thank you @hannesm.
I'll merge now.

PR #74, opened 2024-10-19
Swap around host device and net name

reynir:
The data was swapped according to the table header. That is, the network name internal to the unikernel was displayed under the title "Host Device" while the host bridge name was displayed under "Name".

### DIFF
--- a/unikernel_single.ml
+++ b/unikernel_single.ml
@@ -222,7 +222,7 @@ let unikernel_single_layout ~csrf unikernel now console_output =
                                                     text-primary-600 uppercase";
                                                  ];
                                              ]
-                                           [ txt "Host device" ];
+                                           [ txt "Unikernel block device" ];
                                          th
                                            ~a:
                                              [
@@ -233,7 +233,7 @@ let unikernel_single_layout ~csrf unikernel now console_output =
                                                     text-primary-600 uppercase";
                                                  ];
                                              ]
-                                           [ txt "Name" ];
+                                           [ txt "Host device" ];
                                          th
                                            ~a:
                                              [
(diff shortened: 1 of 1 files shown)


hannesm:
Great observation. I wonder whether we should go a bit deeper here, and use a different wording.

So, we have the name as in the unikernel / solo5 manifest; and we have the name of the host system bridge (i.e. what is in the albatross policy).

Naming is hard, but maybe: "`Unikernel network device` | `Host device`"? I'd also put the unikernel network device first, but I guess this is a question of taste.

hannesm:
note: same for block devices, is swapped around, needs fixing

hannesm:
I went ahead and proposed the changes I had commented above. Please approve/reject

unikernel_single.ml
```diff
@@ -222,7 +222,7 @@ let unikernel_single_layout ~csrf unikernel now console_output =
                                                     text-primary-600 uppercase";
                                                  ];
                                              ]
-                                           [ txt "Host device" ];
+                                           [ txt "Unikernel block device" ];
```
PizieDust, line 225:
```suggestion
                                           [ txt "Block device" ];
```
I wonder if it's better to just say "Block device", since we already know the table contains unikernel information.

unikernel_single.ml
```diff
@@ -325,7 +326,7 @@ let unikernel_single_layout ~csrf unikernel now console_output =
                                                     text-primary-600 uppercase";
                                                  ];
                                              ]
-                                           [ txt "Host device" ];
+                                           [ txt "Unikernel network device" ];
```
PizieDust, line 329:
```suggestion
                                           [ txt "Network device" ];
```

PizieDust (commented):
Just a little suggestion. Please feel free to not approve them if it's better otherwise

hannesm:
The thing is we want to distinguish what the unikernel uses as name vs what the host system uses.... But I'm open for other wording suggestions.

PizieDust:
> The thing is we want to distinguish what the unikernel uses as name vs what the host system uses.... But I'm open for other wording suggestions.

ahh I see. Makes perfect sense now

PizieDust (approved):

PR #76, opened 2024-10-19
etter filtering and table sorting

### LINKED ISSUE #75: Feature: can we have the tables displayed being sortable by each column (maybe with some javascript)
That'd make it easy to view the unikernels by name, but also by cpuid, or assigned memory etc. I'm sure there's some javascript package around that enables sorting ;)

PizieDust:
closes #75 
cc / @hannesm 
You can now click on the table headers to sort by that columns data. 

also the search box now filters across the entire table and not just the names. :) 
If you type 10, it should probably display the unikernels with a CPUID of 10. Basically it filters from data across the entire table.

### DIFF
--- a/assets/main.js
+++ b/assets/main.js
@@ -32,23 +32,18 @@ function getUnikernelName(url) {
 }
 
 function filterData() {
-	var input, filter, table, tr, td, i, txtValue;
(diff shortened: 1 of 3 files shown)


hannesm:
looks great to me, thanks a lot.

Issue #67, opened 2024-10-15
String escaping / cleaning / String.trim


hannesm:
Carefully review and try to remove the amount of sanitization we need to do. Best find a way / API where we don't need to do pre/post cleaning.

Otherwise this may open the door for attackers (when multiple different inputs lead to the same result)...

hannesm:
Addressed in #80. Though there are still some String.trim for the cookies in Middleware.

hannesm:
Now with #81 all of these are gone.

PR #81, opened 2024-10-29
do not String.trim on cookies

hannesm:
this is more a question, whether it is fine to remove these String.trim? #67


PizieDust (approved):
Thank you @hannesm 
I think this covers all the instances of `String.trim` project wide.

PizieDust:
I'll go ahead and merge this

Issue #66, opened 2024-10-15
guard Yojson calls


hannesm:
Lots of yojson calls (even Yojson.Basic.Util.member) may raise exceptions. We should be careful to guard these in a try..catch - esp. since lots of json comes from the user as input, and thus can be as malformed as possible ;)

Best be done by having a separate module through which we call into Yojson.

hannesm:
A lot is addressed in #80 about this.

Some final remarks:
- in unikernel.ml, we have `        match Yojson.Basic.Util.member User_model.csrf_cookie json with` (which may raise)
- we use Yojson.Basic.to_string and from_string, but then also Yojson.Safe.to_string -- should we decide on Basic vs Safe?

hannesm:
done in #82

PR #82, opened 2024-10-29
Json refactorings

hannesm:
These are two commits:
- the first removes the remaining usage of Yojson.Basic.Utils.member
- it also passes the json_dict directly around from extract_csrf_token, leading to fewer indentation

The second commit uses Yojson.Basic.to_string instead of Yojson.Safe.to_string. I don't know exactly why we should use which - but I counted the usages, and there were more Basic usages.

For #66

### DIFF
--- a/albatross_json.ml
+++ b/albatross_json.ml
@@ -177,53 +177,49 @@ let block_device_of_json js =
       | _, _, _ -> Error (`Msg "couldn't decode json"))
   | _ -> Error (`Msg "bad json, either string or assoc")
 
-let policy_of_json js =
-  match js with
-  | `Assoc xs -> (
-      match
-        Utils.Json.
-          ( get "unikernels" xs,
-            get "memory" xs,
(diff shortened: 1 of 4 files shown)


hannesm:
Best reviewed with the `?w=1` added https://github.com/robur-coop/mollymawk/pull/82/files?w=1 -- so that whitespace difference aren't shown.

hannesm:
From a `diff -u basic.ml safe.ml`, I learn that Yojson.Safe.t has additional constructors:
```OCaml
 type t =
     | `Null
     | `Bool of bool
     | `Int of int
+    | `Intlit of string
     | `Float of float
     | `String of string
     | `Assoc of (string * t) list
     | `List of t list
+    | `Tuple of t list
+    | `Variant of (string * t option)
     ]
```

PizieDust (approved):
I think this is some great refactoring. Thank you @hannesm

PR #88, opened 2024-11-05
refactor: use String.starts_with instead of String.length and String.…

hannesm:
…sub and String.equal


PizieDust (approved):
LGTM.
Thank you @hannesm

PR #89, opened 2024-11-05
Refactorings

hannesm:
Best viewed commit by commit, I tried to describe the changes in the commit message.

Sorry that it is a big diff, let me know if we should pair on it. I only tested compilation, not the functionality.


hannesm:
Maybe the `Middleware.redirect_to_page` calls during authentication should be kept? This redirects, clears the session cookie, and also creates such a flash message cookie.

PizieDust:
Thank you @hannesm

PR #94, opened 2024-11-15
Resource Usage

### LINKED ISSUE #83: Provide a view for the user to see the available resources
As administrator, we can see this for all users. As user we should have this on the "My account" page.

PizieDust:
closes #83 
This PR displays the current usage of policy resources for the users account. 

We are using multiple pie charts and bar charts to display this information in a more comprehensible way.

### DIFF
--- a/account_usage.ml
+++ b/account_usage.ml
@@ -24,7 +24,7 @@ let account_usage_layout policy unikernels blocks =
         Hashtbl.replace cpuid_count cpuid (count + 1))
       unikernels;
     (* Prepare list of all cpuids from policy *)
-    let policy_cpuids = Vmm_core.IS.to_list policy.cpuids in
+    let policy_cpuids = Vmm_core.IS.elements policy.cpuids in
     (* Generate the list with all policy.cpuids and their respective counts *)
     List.map
       (fun cpuid ->
--- a/account_usage.ml
+++ b/account_usage.ml
@@ -1,0 +1,362 @@
+let account_usage_layout policy unikernels blocks =
+  let deployed_unikernels = List.length unikernels in
+  let total_volume_used =
+    List.fold_left (fun total_size (_, size, _) -> total_size + size) 0 blocks
+  in
+  let total_free_space =
+    Option.value ~default:0 policy.Vmm_core.Policy.block - total_volume_used
+  in
+  let total_memory_used =
+    List.fold_left
+      (fun total_memory (_, unikernel) ->
+        total_memory + unikernel.Vmm_core.Unikernel.memory)
+      0 unikernels
+  in
+  let count_cpuid_usage =
+    let cpuid_count = Hashtbl.create (List.length unikernels) in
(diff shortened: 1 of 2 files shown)


PizieDust:
@hannesm we have some really bad formatting for the inline javascript. 
At some point in the future we can switch to `js_of_ocaml` :)

account_usage.ml
```diff
@@ -0,0 +1,362 @@
...
+    List.iter
+      (fun (_, unikernel) ->
+        let cpuid = unikernel.Vmm_core.Unikernel.cpuid in
+        let count =
+          Hashtbl.find_opt cpuid_count cpuid |> Option.value ~default:0
+        in
+        Hashtbl.replace cpuid_count cpuid (count + 1))
+      unikernels;
+    (* Prepare list of all cpuids from policy *)
+    let policy_cpuids = Vmm_core.IS.to_list policy.cpuids in
```
hannesm, line 27:
```suggestion
    let policy_cpuids = Vmm_core.IS.elements policy.cpuids in
```

That's the Set.S function to convert the contents of the set to a list.

account_usage.ml
```diff
@@ -0,0 +1,362 @@
...
+
+    (* Convert the hash table to a sorted list of (bridge_name, count) *)
+    let bridge_usage_list =
+      Hashtbl.fold
+        (fun bridge count acc -> (bridge, count) :: acc)
+        bridge_count []
+    in
+    (* Sort the list by bridge name *)
+    List.sort (fun (b1, _) (b2, _) -> String.compare b1 b2) bridge_usage_list
+  in
```
hannesm, line 70: with the above code, I struggle a bit: there's non-trivial counting and lots of data structures involved. Maybe best to put such code in separate functions. I still don't grasp what the data structures for the UI should be? sorted lists?

Especially, I'm not a big fan of hashtable.

All of that shouldn't block this PR from being merged, but we may want to refactor later.

account_usage.ml
```diff
@@ -0,0 +1,362 @@
...
+                          Chart(document.getElementById('cpuChart').getContext('2d'), \
+                          {\n\
+                         \                            type: 'bar',\n\
+                         \                            data: {\n\
+                         \                              labels: [\n\
+                         \                                "
+                        ^ String.concat ", "
+                            (List.map
+                               (fun id -> "\"" ^ string_of_int id ^ "\"")
+                               (Vmm_core.IS.to_list policy.cpuids))
```
hannesm, line 222:
```suggestion
                               (Vmm_core.IS.elements policy.cpuids))
```

account_usage.ml
PizieDust: My idea was to have a chart of the different bridges assigned to the user, and how many times each bridge is used. 
It's the same with cpuid.

If there's a more trivial way to get this information I'll be so happy to refactor

hannesm:
Thanks, I'll merge. I have to see the charts to get an idea which wording would be good. :)

Issue #92, opened 2024-11-13
Volumes UI: "View" is unclear


hannesm:
The volumes UI looks amazing. But I initially was confused by the `View` button -- since I expected to get the data of the volume -- but instead I was redirected to the unikernel that is using the block device.

Since the "get" and "set" operations are to be implemented (see #32), we should find some good names for the buttons :)

I also like the pie chart a lot. Thanks so much @PizieDust :)

PizieDust:
Yes that's right. "View" is not very explanatory. 
I was thinking of a way to display the data. For instance if we read data from a block device, do we just display this as text? What is the data corresponds to images?

Also, the way I understand the "get" operation, we have to give it an offset so it reads to that offset?

hannesm:
> Yes that's right. "View" is not very explanatory. I was thinking of a way to display the data. For instance if we read data from a block device, do we just display this as text? What is the data corresponds to images?

I'd have a "download" button, and allow the data to be downloaded as application/octet-stream.

> Also, the way I understand the "get" operation, we have to give it an offset so it reads to that offset?

The `int` provided to `` `Block_dump `` is the compression level to use. 0 for none, 9 for highest (I guess). It uses https://github.com/mirage/decompress.

PizieDust:
Oh I see. Thank you. I think I understand it now.

hannesm:
#95 addresses this

Issue #32, opened 2024-09-05
Block device management


PizieDust:
Dynamically perform actions on block devices from the dashboard UI. 
These actions will include:

- [x] creating a block device
- [x] dump data into a block device at the time of creation 
- [x] updating a block device
- [x] deleting/destroying a block device
- [x] dumping data into a block device
- [x] backing up data from a block device (downloading)

PR #95, opened 2024-11-20
Upload and download data from and to volumes

### LINKED ISSUE #93: Create volume dialog, error messages
When I enter an invalid name, it displays in red "please enter a name", but it should tell me "invalid name" and maybe the constraints on the name.

PizieDust:
closes #93 #92 #32 

This PR adds the functionality to download data and upload data to and from block devices. 
There's also some refactoring for modals where all is moved to a single component, same with buttons.

At the moment, we can't download a volume if it is in use by a running unikernel, we also can't upload data to a volume which is in use by a running unikernel.


PizieDust:
@hannesm I don't know if this is intended but uploads which are larger in size than the volume are successful. For example uploading a 5MB file to a 1MB volume is successful. I don't know if this should be the case. 
also, empty uploads are successful (uploads with no file).

assets/main.js
```diff
@@ -579,10 +579,32 @@ async function createVolume() {
 	const block_compressed = document.getElementById("block_compressed").checked;
 	const block_data = document.getElementById("block_data").files[0];
 
-	if (!isValidName(block_name)) {
```
hannesm, line 582: Thanks for refining the error message. Now, I'd appreciate if the `deployUnikernel()` as well gets such nice informative error messages, and best have both share the code....

So, could "isValidName" return the error message?

hannesm:
Thanks for your PR:

> @hannesm I don't know if this is intended but uploads which are larger in size than the volume are successful. For example uploading a 5MB file to a 1MB volume is successful.

This is strange, from the albatross source:

```OCaml
  | `Block_set (compressed, data) ->
    begin match Vmm_resources.find_block t.resources id with
      | None -> Error (`Msg "set block: not found")
      | Some (_, true) -> Error (`Msg "set block: is in use")
      | Some (size, false) ->
        let* data =
          if compressed then
            Vmm_compress.uncompress data
          else
            Ok data
        in
        let* size_in_bytes = Vmm_unix.bytes_of_mb size in
        let* () =
          if size_in_bytes >= String.length data then
            Ok ()
          else
            Error (`Msg "data exceeds block size")
        in
        let* () = Vmm_unix.destroy_block id in
        let* () = Vmm_unix.create_block ~data id size in
        Ok (t, `End (`Success (`String "set block device")))
    end
```

So, we first lookup the block name in the resources, which returns us a size (in MB), then we compare that this size (converted to bytes) is greater or equal to the length of the uncompressed data (String.length data). Does, when you uploaded a 5 MB data blob to a 1 MB block device the device size increase to 5 MB? was all data written?

> I don't know if this should be the case. also, empty uploads are successful (uploads with no file).

I don't know exactly what you mean here. So, if albatross receives a "Block_set" with an empty string, it writes all 0s to the block device (which is always truncated to the size that was specified when it was created).

What does mollymawk emit to albatross in the case of "uploads with no file"?

PizieDust:
Both cases emit "set block device" which is a success response.

When I upload a larger file, there's no way to tell how many bytes get
uploaded.

unikernel.ml
```diff
@@ -1328,6 +1328,102 @@ struct
             Middleware.http_response reqd ~title:"Error"
               ~data:"Couldn't find fields" `Bad_request)
 
+  let download_volume json_dict albatross reqd (user : User_model.user) =
+    match
+      Utils.Json.(get "block_name" json_dict, get "compression_level" json_dict)
+    with
+    | Some (`String block_name), Some (`Int compression_level) -> (
```
hannesm, line 1335: here we need a different interface for albatross, and have some streaming or chunk read & write - likely with a chunked transfer-encoding.

unikernel.ml
```diff
@@ -1328,6 +1328,102 @@ struct
...
+            | Error (`String res) ->
+                Middleware.http_response reqd ~title:"Error"
+                  ~data:(Yojson.Basic.to_string (`String res))
+                  `Internal_server_error))
+    | _ ->
+        Middleware.http_response reqd ~title:"Error"
+          ~data:"Couldn't find block name in json" `Bad_request
+
+  let upload_to_volume albatross reqd (user : User_model.user) =
+    read_multipart_data reqd >>= fun result ->
```
hannesm, line 1363: similar to unikernel_deploy, we should (at one day in our live) have a streaming interface here as well, instead of requiring the entire data fitting into memory.

the downside (and we'll likely need to revise the protocol) is that in TLS we can embed only 16MB of certificates, this means that we can't write/set multi-gigabyte block devices. for now, this is a shortcoming -- I opened an issue at the albatross repository https://github.com/robur-coop/albatross/issues/201

unikernel.ml
```diff
@@ -1328,6 +1328,102 @@ struct
...
+    match result with
```
hannesm, line 1364:
```suggestion
    read_multipart_data reqd >>= function
```

PR #96, opened 2024-12-02
Token Management

PizieDust:
related to #38 
This PR begins the work for allowing accessing mollymawk solely via it's API.

API tokens are no longer created when a user creates an account. 
The user can manually create an API token from their dashboard. 

The API tokens now have new fields:
- `name` : to help the user differentiate between different keys e.g (GitHub, GitLab)
- `last_access`: when last was the specific key used to query mollymawk
- `usage_count`: how many times this key has been used

### DIFF
--- a/user_model.ml
+++ b/user_model.ml
@@ -4,7 +4,8 @@ type token = {
   name : string;
   token_type : string;
   value : string;
-  expires_in : int; (* the number of seconds until this token is invalid, starts counting from created_at*)
+  expires_in : int;
+      (* the number of seconds until this token is invalid, starts counting from created_at*)
   created_at : Ptime.t;
   last_access : Ptime.t;
   usage_count : int;
--- a/user_model.ml
+++ b/user_model.ml
@@ -1,10 +1,13 @@
 module Rng = Mirage_crypto_rng
 
 type token = {
+  name : string;
   token_type : string;
   value : string;
   expires_in : int;
   created_at : Ptime.t;
+  last_access : Ptime.t;
+  usage_count : int;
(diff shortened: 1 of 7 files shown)


user_model.ml
```diff
@@ -1,10 +1,13 @@
...
 
 type token = {
+  name : string;
   token_type : string;
   value : string;
   expires_in : int;
   created_at : Ptime.t;
+  last_access : Ptime.t;
+  usage_count : int;
       (* In seconds, so after 1 hour would be 3600 seconds of inactivity *)
```
hannesm, line 11: I don't understand the comment here...

What I see is a created_at and expires_in, so I'd assume it is valid from created_at until created_at + expires_in (in seconds?).

Would be great to have that documented. the comment here is a bit off/unattached now.

user_model.ml
```diff
@@ -1,10 +1,13 @@
 module Rng = Mirage_crypto_rng
 
 type token = {
+  name : string;
   token_type : string;
```
hannesm, line 5: the token type is always "Bearer"?

hannesm:
looks fine to me, please merge when you think it is ready

user_model.ml
PizieDust: This comment slipped in. I don't think it's relevant anymore.

user_model.ml
PizieDust: At the moment yes, it's always `Bearer` but I like that it's flexible and we can always change the authentication type. maybe `Basic` someday?

PizieDust:
Thank you @hannesm for the review.

PR #97, opened 2024-12-08
Api authentication with tokens

### LINKED ISSUE #38: REST API

PizieDust:
closes #38 

This PR introduces many changes, including a refactoring of how we check for `csrf` tokens. 
In this PR:
- We can now have both session authentication and api token authentication (both stateful and stateless)
- Routes which change state are all converted to POST requests so that we can do the appropriate verifications.

The following endpoints can be accessed via the API
- [x] `/api/volume/delete`
- [x] `/api/volume/create`
- [x] `/api/volume/download`
- [x] `/api/volume/upload`
- [x] `/api/unikernel/destroy`
- [x] `/api/unikernel/restart`
- [x] `/api/unikernel/create`

### DIFF
--- a/middleware.ml
+++ b/middleware.ml
@@ -101,7 +101,8 @@ let redirect_to_dashboard reqd ?(msg = "") () =
   Httpaf.Reqd.respond_with_string reqd response msg;
   Lwt.return_unit
 
-let http_response ~title ?(header_list = []) ?(data = "") reqd http_status =
+let http_response ~title ?(header_list = []) ?(data = `String "") reqd
+    http_status =
   let code = Httpaf.Status.to_code http_status
   and success = Httpaf.Status.is_successful http_status in
   let status = { Utils.Status.code; title; data; success } in
@@ -152,7 +153,9 @@ let is_user_admin_middleware api_meth user handler reqd =
   if user.User_model.super_user && user.active then handler reqd
   else
     redirect_to_error ~title:"Unauthorized"
-      ~data:"You don't have the necessary permissions to access this service."
+      ~data:
+        (`String
+          "You don't have the necessary permissions to access this service.")
       `Unauthorized 401 api_meth reqd ()
 
 let csrf_cookie_verification form_csrf reqd =
@@ -173,10 +176,12 @@ let csrf_verification user now form_csrf handler reqd =
       if User_model.is_valid_cookie csrf_token now then handler reqd
       else
         http_response ~title:"CSRF Token Mismatch"
-          ~data:"Invalid CSRF token error. Please refresh and try again." reqd
-          `Bad_request
+          ~data:
+            (`String "Invalid CSRF token error. Please refresh and try again.")
+          reqd `Bad_request
   | None ->
-      http_response ~data:"Missing CSRF token. Please refresh and try again."
+      http_response
+        ~data:(`String "Missing CSRF token. Please refresh and try again.")
         ~title:"Missing CSRF Token" reqd `Bad_request
 
 let api_authentication reqd =
--- a/middleware.ml
+++ b/middleware.ml
@@ -178,3 +178,10 @@ let csrf_verification user now form_csrf handler reqd =
   | None ->
       http_response ~data:"Missing CSRF token. Please refresh and try again."
         ~title:"Missing CSRF Token" reqd `Bad_request
+
+let api_authentication reqd =
+  match header "Authorization" reqd with
+  | Some auth when String.starts_with ~prefix:"Bearer " auth ->
+      let token = String.sub auth 7 (String.length auth - 7) in
+      Some token
+  | _ -> None
--- a/middleware.ml
+++ b/middleware.ml
@@ -139,24 +139,9 @@ let session_cookie_value reqd =
           m "auth-middleware: No molly-session in cookie header.");
       Error (`Msg "User not found")
 
-let auth_middleware user handler reqd =
-  if user.User_model.active then handler reqd
-  else
-    redirect_to_page ~path:"/sign-in" ~clear_session:true ~with_error:true
-      ~msg:"User account is deactivated." reqd ()
-
 let email_verified_middleware user handler reqd =
   if User_model.is_email_verified user then handler reqd
   else redirect_to_verify_email reqd ()
-
-let is_user_admin_middleware api_meth user handler reqd =
-  if user.User_model.super_user && user.active then handler reqd
-  else
-    redirect_to_error ~title:"Unauthorized"
-      ~data:
-        (`String
-          "You don't have the necessary permissions to access this service.")
-      `Unauthorized 401 api_meth reqd ()
 
 let csrf_cookie_verification form_csrf reqd =
   match cookie User_model.csrf_cookie reqd with
--- a/unikernel.ml
+++ b/unikernel.ml
@@ -1804,7 +1804,8 @@ struct
                   (delete_volume !albatross reqd))
         | "/api/volume/create" ->
             check_meth `POST (fun () ->
-                authenticate ~check_token:true ~check_csrf:true ~api_meth:true store reqd
+                authenticate ~check_token:true ~check_csrf:true ~api_meth:true
+                  store reqd
                   (create_volume !albatross reqd))
         | "/api/volume/download" ->
             check_meth `POST (fun () ->
@@ -1813,7 +1814,8 @@ struct
                   (download_volume !albatross reqd))
         | "/api/volume/upload" ->
             check_meth `POST (fun () ->
-                authenticate ~check_token:true ~check_csrf:true ~api_meth:true store reqd
+                authenticate ~check_token:true ~check_csrf:true ~api_meth:true
+                  store reqd
                   (upload_to_volume !albatross reqd))
         | "/tokens" ->
             check_meth `GET (fun () ->
@@ -1889,7 +1891,8 @@ struct
                   (unikernel_destroy !albatross reqd))
         | "/api/unikernel/restart" ->
             check_meth `POST (fun () ->
-                authenticate ~check_token:true ~check_csrf:true ~api_meth:true store reqd
+                authenticate ~check_token:true ~check_csrf:true ~api_meth:true
+                  store reqd
                   (unikernel_restart !albatross reqd))
         | path when String.starts_with ~prefix:"/unikernel/console/" path ->
             check_meth `GET (fun () ->
@@ -1900,7 +1903,8 @@ struct
                   (unikernel_console !albatross unikernel_name reqd))
         | "/api/unikernel/create" ->
             check_meth `POST (fun () ->
-                authenticate ~check_token:true ~check_csrf:true ~api_meth:true store reqd
+                authenticate ~check_token:true ~check_csrf:true ~api_meth:true
+                  store reqd
                   (unikernel_create !albatross reqd))
         | _ ->
             let error =
--- a/unikernel.ml
+++ b/unikernel.ml
@@ -1889,7 +1889,7 @@ struct
                   (unikernel_destroy !albatross reqd))
         | "/api/unikernel/restart" ->
             check_meth `POST (fun () ->
-                authenticate ~check_token:true ~check_csrf:true store reqd
+                authenticate ~check_token:true ~check_csrf:true ~api_meth:true store reqd
                   (unikernel_restart !albatross reqd))
         | path when String.starts_with ~prefix:"/unikernel/console/" path ->
             check_meth `GET (fun () ->
--- a/unikernel.ml
+++ b/unikernel.ml
@@ -217,9 +217,10 @@ struct
           `Not_found
 
   let authenticate ?(email_verified = true) ?(check_admin = false)
-      ?(api_meth = false) ?(check_csrf = false) store reqd f =
+      ?(api_meth = false) ?(check_csrf = false) ?(check_token = false) store
+      reqd f =
     let current_time = Ptime.v (P.now_d_ps ()) in
-    match (Middleware.api_authentication reqd, api_meth) with
+    match (Middleware.api_authentication reqd, check_token) with
     | Some token_value, true -> (
         extract_json_body reqd >>= function
         | Ok json_dict ->
@@ -229,6 +230,13 @@ struct
             Middleware.http_response reqd ~title:"Error"
               ~data:(`String (String.escaped msg))
               `Bad_request)
+    | Some _token_value, false ->
+        Middleware.http_response reqd ~title:"Error"
+          ~data:
+            (`String
+              "This endpoint cannot be accessed via the API. Please use the \
+               web dashboard.")
+          `Bad_request
     | _ ->
         if check_csrf then
           extract_csrf_token reqd >>= function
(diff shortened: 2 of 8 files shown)


middleware.ml
```diff
@@ -101,7 +101,8 @@ let redirect_to_dashboard reqd ?(msg = "") () =
   Httpaf.Reqd.respond_with_string reqd response msg;
   Lwt.return_unit
 
-let http_response ~title ?(header_list = []) ?(data = "") reqd http_status =
+let http_response ~title ?(header_list = []) ?(data = `String "") reqd
```
hannesm, line 104: What is the reasoning behind this change, putting `data` as some json object instead of a string?

I'm asking since this change adds a lot of changes to this PR, and I'm not able to figure out why that needs to change?

I was able to find this code:
```Diff
                        Middleware.http_response reqd ~header_list
                          ~title:"Success"
                          ~data:
                            (Yojson.Basic.to_string
                               (User_model.user_to_json user))
                          ~data:(User_model.user_to_json user)
                          `OK
```

But am still unsure why this json object is now necessary.

unikernel.ml
```diff
@@ -128,66 +142,128 @@ struct
...
+      @ [ Middleware.auth_middleware user ]
+    in
+    let handle_csrf () =
+      if check_csrf then
+        extract_csrf_token reqd >>= function
+        | Ok (form_csrf, json_dict) ->
+            Lwt.return (Ok (Some (form_csrf, json_dict)))
+        | Error (`Msg msg) -> Lwt.return (Error (`Msg msg))
+      else Lwt.return (Ok None)
+    in
```
hannesm, line 165: can we please pu the `check_csrf` directly where this function is being called

unikernel.ml
```diff
@@ -128,66 +142,128 @@ struct
...
                   Store.update_user store updated_user >>= function
-                  | Ok () -> f user
+                  | Ok () -> f ~json_dict user
                   | Error (`Msg err) ->
                       Logs.err (fun m -> m "Error with storage: %s" err);
-                      Middleware.http_response reqd ~title:"Error" ~data:err
-                        `Not_found)
-                reqd)
+                      Middleware.http_response reqd ~title:"Error"
+                        ~data:(`String err) `Not_found)
```
hannesm, line 239: can this please be moved to a function in storage (or parts in user_model)? I don't think in `authenticate` we should think about "how to increment a token counter"...

unikernel.ml
```diff
@@ -751,7 +840,37 @@ struct
                 ~icon:"/images/robur.png" ())
              `Unauthorized)
 
-  let users store reqd (user : User_model.user) =
+  let close_session store reqd ~json_dict (user : User_model.user) =
```
hannesm, line 843: I struggle with `close_sessions` vs `close_session` -- what is the idea behind having two separate functions?

unikernel.ml
```diff
@@ -1810,41 +1859,31 @@ struct
         | "/unikernel/deploy" ->
             check_meth `GET (fun () ->
                 authenticate store reqd (deploy_form store reqd))
-        | "/unikernel/destroy" ->
+        | "api/unikernel/destroy" ->
```
hannesm, line 1862: why is there suddenly no leading `/`?

hannesm:
One question I have is: which endpoints should be accessible with an API token? All of them? As many as possible? Or is there a plan / strategy which endpoints?

I.e. does a "change password" endpoint make sense to be available with a token? The underlying question is as well what permission does a token have / should a user when creating a token assign permissions to it? These are open design questions.

unikernel.ml
PizieDust: so `close_sessions` just uses the users session cookie and deletes every other cookie except the one making the request while `close_sesssion` has a json body in the request with the specific cookie the user wants to delete.

unikernel.ml
PizieDust: this should be an error. I'll fix that in a commit

middleware.ml
PizieDust: So the reason why `data` is now a json object is because we use it for both responses to the API and also responses to the UI. When we format data as a string, the api responses also return with data as a string. When we make data an object, we can have the API responses return correctly. This is an example of the response when creating a new token via the API.

Without `data` as a json object this is what the results come as:

```json
{
    "status": 200,
    "title": "Success",
    "success": true,
    "data": {
    "status": 200,
    "title": "Success",
    "success": true,
    "data": "{\"token_type\":\"Bearer\",\"value\":\"7888febc-fb8c-4648-acb8-e449ba27ee88\",\"expires_in\":3600,\"created_at\":\"2024-12-09 11:46:01-00:00\",\"last_access\":\"2024-12-09 11:46:01-00:00\",\"name\":\"api_test_token\",\"usage_count\":0}"
}
}
```

with data as a json object, we can have this, and then for results of the UI such as in notifications we can do it as a string:
```json
{
    "status": 200,
    "title": "Success",
    "success": true,
    "data": {
        "token_type": "Bearer",
        "value": "6b8fe160-1dbd-4193-a5f8-8459d2c5d59a",
        "expires_in": 3600,
        "created_at": "2024-12-09 11:39:59-00:00",
        "last_access": "2024-12-09 11:39:59-00:00",
        "name": "api_test_token",
        "usage_count": 0
    }
}
```

PizieDust:
> One question I have is: which endpoints should be accessible with an API token? All of them? As many as possible? Or is there a plan / strategy which endpoints?
> 
> I.e. does a "change password" endpoint make sense to be available with a token? The underlying question is as well what permission does a token have / should a user when creating a token assign permissions to it? These are open design questions.

I think the endpoints that should be accessible with the token should be endpoints that do anything with albatross (creating a unikernel, volumes, destroying them etc) and maybe requesting new tokens. 

For other activities, I think these should only be done via the dashboard

unikernel.ml
```diff
@@ -368,32 +446,34 @@ struct
...
+                        "CSRF token mismatch error. Please referesh and try \
+                         again.") `Bad_request)
         | _ ->
             Middleware.http_response reqd ~title:"Error"
               ~data:
-                (Fmt.str "Register: Unexpected fields. Got %s"
-                   (Yojson.Basic.to_string (`Assoc json_dict)))
+                (`String
+                  (Fmt.str "Register: Unexpected fields. Got %s"
+                     (Yojson.Basic.to_string (`Assoc json_dict))))
```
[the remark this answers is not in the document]
PizieDust: Here it's encoded as a string because it's going to the UI to be displayed in a toast notification.
In some other cases, particularly that which is only an api response, we want to return the json object as is.

unikernel.ml
```diff
@@ -999,35 +1109,35 @@ struct
...
                         | Ok res ->
                             Middleware.http_response reqd ~title:"Success"
-                              ~data:(Yojson.Basic.to_string res)
-                              `OK
+                              ~data:res `OK
                         | Error (`String res) ->
                             Middleware.http_response reqd ~title:"Error"
-                              ~data:(Yojson.Basic.to_string (`String res))
-                              `Internal_server_error))
+                              ~data:(`String res) `Internal_server_error))
```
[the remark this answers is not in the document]
PizieDust: Yes I think it's better to escape the string before returning. I'll have a commit for that

PizieDust:
> One question I have is: which endpoints should be accessible with an API token? All of them? As many as possible? Or is there a plan / strategy which endpoints?
> 
> I.e. does a "change password" endpoint make sense to be available with a token? The underlying question is as well what permission does a token have / should a user when creating a token assign permissions to it? These are open design questions.

We can add permissions to the tokens, I think that's a great idea, I can have another PR with that

hannesm:
> I think the endpoints that should be accessible with the token should be endpoints that do anything with albatross (creating a unikernel, volumes, destroying them etc) and maybe requesting new tokens.

I completely agree, and would appreciate if we add - in this PR - a flag for authenticate `~token_is_ok:bool`. Please let us _not_ dive into the potential permission implementations.

I would not add `requesting new tokens` to the set of endpoints that can be done with a token.

To be more explicit, I'd guess the following endpoints are fine with a token:
- /api/volume/delete
- /api/volume/create
- /api/volume/download
- /api/volume/upload
- /api/unikernel/destroy
- /api/unikernel/restart
- /api/unikernel/create

Would you agree? Is there something missing?

PizieDust:
> > I think the endpoints that should be accessible with the token should be endpoints that do anything with albatross (creating a unikernel, volumes, destroying them etc) and maybe requesting new tokens.
> 
> I completely agree, and would appreciate if we add - in this PR - a flag for authenticate `~token_is_ok:bool`. Please let us _not_ dive into the potential permission implementations.
> 
> I would not add `requesting new tokens` to the set of endpoints that can be done with a token.
> 
> To be more explicit, I'd guess the following endpoints are fine with a token:
> 
> * /api/volume/delete
> * /api/volume/create
> * /api/volume/download
> * /api/volume/upload
> * /api/unikernel/destroy
> * /api/unikernel/restart
> * /api/unikernel/create
> 
> Would you agree? Is there something missing?

This sounds great. I definitely agree

unikernel.ml
hannesm: I still don't understand why there used to be one function and now we have multiple functions.

PizieDust: so previously one of the functions `close_session` was a `GET` request and when doing this token authentication it was unclear to me at the time which of the requests will go via the API, so I changed this into a `POST` request and now the cookie is sent in the json body.  Given that the endpoints are more clearly defined now I think this change can be reverted

Given that

unikernel.ml
```diff
@@ -155,40 +180,77 @@ struct
...
+        Middleware.http_response reqd ~title:"Error"
+          ~data:
+            (`String
+              ("Invalid authorization token. User not found for token "
+             ^ token_value))
+          `Not_found
+
+  let authenticate ?(email_verified = true) ?(check_admin = false)
+      ?(api_meth = false) ?(check_csrf = false) ?(check_token = false) store
+      reqd f =
```
hannesm, line 221: I struggle with the logic below - esp. given the 5 optional boolean parameters.

Would it make sense to write it in english what should be done in this function? An attempt:

"Trying to access any endpoint, we need to check user permissions. Depending on the endpoint, the user may authenticate with a cookie or with a token (API request).

So, the main logic is:
  if token_is_sufficient then
    if check_for_token then Ok `Token else if check_cookie then Ok `Cookie else Error ()
  else
    if check_cookie then Ok `Cookie else Error ()

Now, depending on other parameters, more checks need to be done: CSRF if check_csrf is true and Ok `Cookie is used.

Also, check_admin must be respected for the endpoints requiring higher privileges.

The api_meth is used ???, email_verified is used ???"

But here, I see a first match on "api_authentication" and "check_token", where additional nesting is done (extract_json_body). In a different branch we get an error, in another branch we have conditionals, and further arguments are passed along....

It feels more like an `authentication_and_process_request_and_decode_json_and_check_for_privileges_and_maybe_use_token_authentication_and_error_with_either_text_plain_or_json` -- so I'd guess we need more declarative code: do one thing in a function, do not overload with lots of optional parameters (that tend to be forgotten), and compose these smaller functions.

hannesm: Exposure: consider we have a malicious person in the room who tries to get something with mollymawk. Now, to minimize the attack surface, we should avoid doing lots of user input handling before we verified that the request is auhenticated.

The reasoning: dealing with user input is hairy, and may contain bugs (leading to exceptions etc.). If we expose the input handling (e.g. json parsing) to a point when the user is auhenticated, 99.999% of attacks will vanish in the void since the request was not from an authenticated source.

PizieDust: So at the moment we don't use `email_verified` is not used, since we don't yet send any verification emails. 

`check_admin` as you correctly mentioned is used to check if the user can access that endpoint.

We have methods that return a UI (such as opening the login page) and the rest of the methods return JSON (such as processing the login request). To differentiate between these two types of endpoints, we are using:
`api_meth` for those that return JSON and otherwise we assume this should return UI.

`check_csrf` is used to indicate which routes must contain a csrf token and `check_token` is used to indicate which routes are accessible via the API interface.

Without the `check_csrf` flag, most GET request will fail as these don't have any csrf in the request and without `check_token`, everything is accessible from the API.

I think these are all the labels. 

So what `authenticate` is currently doing is checking if there's an authorization header in the request and if `check_token` is true, this way we know this request needs to be processed as an API request and we then extract the request JSON body and process the request. If we do have an authorization header and `check_token` is false, it could mean the user is trying to access another route (maybe create a new token) and here we can say the route is inaccessible via the api,

Now if there's no Authorization header and check_token is false, we can assume this request is coming from the UI and here we first `extract_csrf_token` and then process the request.

PizieDust: > Exposure: consider we have a malicious person in the room who tries to get something with mollymawk. Now, to minimize the attack surface, we should avoid doing lots of user input handling before we verified that the request is auhenticated.
> 
> The reasoning: dealing with user input is hairy, and may contain bugs (leading to exceptions etc.). If we expose the input handling (e.g. json parsing) to a point when the user is auhenticated, 99.999% of attacks will vanish in the void since the request was not from an authenticated source.

Thank you for this. I think you are absolutely correct. We should perform all the authentication steps before actually looking at what the request wants. I'll do some refactoring and ping you

unikernel.ml
```diff
@@ -1642,101 +1780,56 @@ struct
...
-                | Error (`Msg msg) ->
-                    Middleware.http_response reqd ~title:"Error"
-                      ~data:(String.escaped msg) `Bad_request)
+                authenticate ~check_token:true ~check_csrf:true ~api_meth:true
+                  store reqd
+                  (delete_volume !albatross reqd))
         | "/api/volume/create" ->
             check_meth `POST (fun () ->
-                authenticate store reqd (create_volume !albatross reqd))
+                authenticate ~check_token:true ~check_csrf:true store reqd
```
hannesm, line 1807: shouldn't there be `~api_meth:true`?
```suggestion
                authenticate ~check_token:true ~check_csrf:true ~api_meth:true store reqd
```

unikernel.ml
```diff
@@ -1642,101 +1780,56 @@ struct
...
-                | Error (`Msg msg) ->
-                    Middleware.http_response reqd ~title:"Error"
-                      ~data:(String.escaped msg) `Bad_request)
+                authenticate ~check_token:true ~check_csrf:true ~api_meth:true
+                  store reqd
+                  (download_volume !albatross reqd))
         | "/api/volume/upload" ->
             check_meth `POST (fun () ->
-                authenticate store reqd (upload_to_volume !albatross reqd))
+                authenticate ~check_token:true ~check_csrf:true store reqd
```
hannesm, line 1816:
```suggestion
                authenticate ~check_token:true ~check_csrf:true ~api_meth:true store reqd
```

here as well

unikernel.ml
```diff
@@ -1810,41 +1883,32 @@ struct
...
+        | "/api/unikernel/restart" ->
             check_meth `POST (fun () ->
-                extract_csrf_token reqd >>= function
-                | Ok (form_csrf, json_dict) ->
-                    authenticate store reqd ~form_csrf
-                      (unikernel_restart json_dict !albatross reqd)
-                | Error (`Msg msg) ->
-                    Middleware.http_response reqd ~title:"Error"
-                      ~data:(String.escaped msg) `Bad_request)
+                authenticate ~check_token:true ~check_csrf:true store reqd
```
hannesm, line 1892:
```suggestion
                authenticate ~check_token:true ~check_csrf:true ~api_meth:true store reqd
```

unikernel.ml
```diff
@@ -1810,41 +1883,32 @@ struct
...
                 let unikernel_name =
                   String.sub path 19 (String.length path - 19)
                 in
                 authenticate store reqd
                   (unikernel_console !albatross unikernel_name reqd))
-        | "/unikernel/create" ->
+        | "/api/unikernel/create" ->
             check_meth `POST (fun () ->
-                authenticate store reqd (unikernel_create !albatross reqd))
+                authenticate ~check_token:true ~check_csrf:true store reqd
```
hannesm, line 1903:
```suggestion
                authenticate ~check_token:true ~check_csrf:true ~api_meth:true store reqd
```

Issue #54, opened 2024-09-23
Update a unikernel from our reproducible build database


hannesm:
In albatross-client, we support the `update` command, which (a) fetches from builds.robur.coop the current unikernel job, (b) looks whether there's a more recent build of the same job, and (c) displays a url to the diff, finally (d) deploys the new unikernel

For mollymawk, we should have a two-stage interface where we can click on update, and then the diff is displayed inline, and we can confirm or reject the update.

PizieDust:
@hannesm   I see that `builder-web` uses dream. Do you think it's a great idea to add two new endpoints that just return json data. I think this will be great for mollymawk, this way we just get the data and display it however we want. 
The endpoints I propose should be:

`GET` `.../latest/uuid` -> return the uuid of the latest binary
`GET` `.../compare/json/old_uuid/new_uuid` -> returns differences formatted in json

Let me know what you think.

hannesm:
done in #102 and #105 \o/

PR #102, opened 2025-01-09
Update running unikernel: Get build info from builds.robur.coop and display diff

PizieDust:
Related to https://github.com/robur-coop/mollymawk/issues/54

This PR checks if there is an update to a unikernel from `builds.robur.coop` and if there's an update it displays the differences between the current unikernel and the update. 
At the moment, this PR only displays the diff and doesn't update the unikernel binary (yet).


hannesm:
The PR title is confusing.

unikernel.ml
```diff
@@ -301,6 +301,21 @@ struct
               reply);
         []
 
+  let single_unikernel albatross ~user_name ~unikernel_name =
```
hannesm, line 304: could we name this differently? I'm a bit confused by "single_unikernel" and "unikernel_info_one".

unikernel.ml
```diff
@@ -301,6 +301,21 @@ struct
...
+    Albatross.query albatross ~domain:user_name ~name:unikernel_name
+      (`Unikernel_cmd `Unikernel_info)
+    >|= function
+    | Error msg ->
+        Logs.err (fun m -> m "error while communicating with albatross: %s" msg);
+        []
+    | Ok (_hdr, `Success (`Unikernel_info unikernels)) -> unikernels
```
hannesm, line 311: What about directly here constraining to a single unikernel... and return `(Name.t * Unikernel.info) option` instead of having to deal with that at the call sites?

unikernel.ml
```diff
@@ -1030,6 +1032,142 @@ struct
...
+      let job_response_body = ref "" in
+      Utils.Builder_web.send_request
+        (Utils.Builder_web.base_url ^ "/hash?sha256="
+        ^ Ohex.encode unikernel.digest)
+        job_response_body
+      >>= function
+      | Error (`Msg err) ->
+          Middleware.redirect_to_error
+            ~data:(`String ("Builder_web request: " ^ err))
+            ~title:"Update Error" ~api_meth:false `Internal_server_error reqd ()
```
hannesm, line 1051: Should we be more specific than "Update error"?

unikernel.ml
```diff
@@ -1030,6 +1032,142 @@ struct
...
+      | Ok response_body -> (
+          match
+            Utils.Builder_web.build_of_json
+              (Yojson.Basic.from_string !response_body)
+          with
+          | Error (`Msg err) ->
+              Middleware.redirect_to_error
+                ~data:(`String ("Builder_web request: " ^ err))
+                ~title:"Update Error" ~api_meth:false `Internal_server_error
```
hannesm, line 1060: Same as above. It would help to provide some context.

unikernel.ml
```diff
@@ -1030,6 +1032,142 @@ struct
...
+              let job_response_body = ref "" in
+              Utils.Builder_web.send_request
+                (Utils.Builder_web.base_url ^ "/job/" ^ current_job_data.job
+               ^ "/build/latest")
+                job_response_body
+              >>= function
+              | Error (`Msg err) ->
+                  Middleware.redirect_to_error
+                    ~data:(`String ("Builder_web request: " ^ err))
+                    ~title:"Update Error" ~api_meth:false `Internal_server_error
```
hannesm, line 1072: see above

unikernel.ml
```diff
@@ -1030,6 +1032,142 @@ struct
...
+                      Middleware.redirect_to_error
+                        ~data:(`String ("Builder_web request: " ^ err))
+                        ~title:"Update Error" ~api_meth:false
+                        `Internal_server_error reqd ()
+                  | Ok latest_job_data -> (
+                      match
+                        String.equal latest_job_data.uuid current_job_data.uuid
+                      with
+                      | true ->
+                          Logs.warn (fun m ->
```
hannesm, line 1089: not sure about the log level, warn feels a bit strange.

unikernel_single.ml
```diff
@@ -53,8 +53,27 @@ let unikernel_single_layout unikernel current_time console_output =
...
+                          ];
+                        div
+                          [
+                            a
+                              ~a:
+                                [
+                                  a_href
+                                    ("/unikernel/update/"
+                                    ^ Option.value ~default:""
+                                        (Vmm_core.Name.name u_name));
```
hannesm, line 67: I don't quite understand this `Option` here -- how can a `unikernel_single` be done where there's `u_name` being `None`?

utils.ml
```diff
@@ -93,6 +97,350 @@ module Status = struct
     |> Yojson.Basic.to_string
 end
 
+module Builder_web = struct
```
hannesm, line 100: should that better be in a separate file -- builder_web.ml?

utils.ml
```diff
@@ -93,6 +97,350 @@ module Status = struct
...
+  }
+
+  type o_diff = {
+    package_version : string;
+    effectively_equal : bool;
+    diff : string;
+  }
+
+  type duniverse = { name : string; value_ : string }
+  type duniverse_detailed_diff = { name : string }
```
hannesm, line 128: that's pretty verbose -- can we remove this type (and use `string` instead)?

utils.ml
```diff
@@ -93,6 +97,350 @@ module Status = struct
...
+    left_packages : package list;
+    right_packages : package list;
+    different_versions : package_version_diff list;
+  }
+
+  type duniverse_diff = {
+    left : duniverse list;
+    right : duniverse list;
+    detailed_diff : duniverse_detailed_diff list;
+  }
```
hannesm, line 140: I don't quite understand why these type definitions differ from builder-web -- are the types in builder-web not ergonomic enough? should we separate the opamdiff subpackage into a custom opam package and depend on that in mollymawk (to share the type definitions and avoid code duplication)?

utils.ml
```diff
@@ -93,6 +97,350 @@ module Status = struct
...
+    >>= function
+    | Error (`Msg err) -> Lwt.return (Error (`Msg err))
+    | Ok (resp, ()) -> (
+        match resp.Http_lwt_client.status with
+        | `OK -> Lwt.return (Ok response_body)
+        | _ ->
+            Lwt.return
+              (Error
+                 (`Msg (Format.asprintf "%a" Http_lwt_client.pp_response resp)))
+        )
```
hannesm, line 441:
```suggestion
  let send_request url =
    let body = "" in
    let body_f _ acc chunk =
      Lwt.return (acc ^ chunk)
    in
    Http_lwt_client.request ~follow_redirect:true
      ~headers:[ ("Accept", "application/json") ]
      url body_f body
    >>= function
    | Error (`Msg err) -> Lwt.return (Error (`Msg err))
    | Ok (resp, body) -> (
        match resp.Http_lwt_client.status with
        | `OK -> Lwt.return (Ok body)
        | _ ->
            Lwt.return
              (Error
                 (`Msg (Format.asprintf "%a" Http_lwt_client.pp_response resp)))
        )
```

this way we avoid the reference cell, and deliver the body as output

utils.ml
```diff
@@ -163,3 +511,7 @@ let display_banner = function
               ~btn_type:`Primary_outlined ();
           ])
   | None -> Tyxml_html.div []
+
+let bytes_to_megabytes bytes =
+  let megabytes = float_of_int bytes /. (1024.0 *. 1024.0) in
+  Printf.sprintf "%.2f" megabytes
```
hannesm, line 517: should we write the "MB" as well to the string?

unikernel.ml
PizieDust: Since fetching all unikernels is named `user_unikernels`, I renamed the function to just `user_unikernel` to return a single unikernel from albatross

unikernel_single.ml
PizieDust: You are right. The unikernel_single won't wwork with u_name being empty. 
However, the url endpoint takes just the unikernel name e.g `roburtls` and not `robur:roburtls`. 
The method `Vmm_core.Name.name` function gives a `string option` with explains why `Option.value` is used here and in other places.

PizieDust: But we can update the url endpoints and methods to work with the full unikernel names i.e `robur:roburtls` rather than just `roburtls`.

what do you think?

utils.ml
PizieDust: Yes this will be great to share the code in a package. I think it will ensure consistency now and also in the future.

unikernel_single.ml
hannesm: I still don't quite understand that. The proposal "work with the full unikernel names i.e robur:roburtls" isn't too convincing, since the domain (or how you call it -- the "robur" part) comes via the access permissions / users. If we use a different mechanism, this means we've to implement some logic which mollymawk user has which permissions on which unikernels (albatross domain).

assets/main.js
```diff
@@ -933,3 +956,4 @@ function areCharactersValid(s) {
 	}
 	return true;
 }
+
```
hannesm, line 959: not sure about this newline.

unikernel.ml
```diff
@@ -301,6 +302,27 @@ struct
...
+    >|= function
+    | Error err ->
+        Logs.err (fun m ->
+            m
+              "Error while communicating with albatross. Trying to fetch %s \
+               resulted in : %s"
+              unikernel_name err);
+        None
+    | Ok (_hdr, `Success (`Unikernel_info unikernels)) ->
+        if List.length unikernels > 0 then Some (List.hd unikernels) else None
```
hannesm, line 317:
```suggestion
    | Ok (_hdr, `Success (`Unikernel_info [ unikernel ])) ->
        Some unikernel
   | Ok (_hdr, `Success (`Unikernel_info unikernels)) ->
       Logs.err (fun m -> m "expected a single unikernel information from albatross, received %u" (List.length unikernels));
       None
```

unikernel.ml
```diff
@@ -973,62 +995,189 @@ struct
...
+          (reply reqd ~content_type:"text/html"
+             (Guest_layout.guest_layout ~page_title:"500 | Mollymawk"
+                ~content:
+                  (Error_page.error_layout
+                     {
+                       code = 500;
+                       success = false;
+                       title = "Error fetching " ^ name;
+                       data =
+                         `String ("An error occured trying to fetch " ^ name);
```
hannesm, line 1012: Shouldn't that refer to "albatross" to some degree? The "Error fetching " is pretty mysterious.

unikernel.ml
```diff
@@ -973,62 +995,189 @@ struct
...
+
+  let unikernel_prepare_update albatross store name reqd http_client
+      ~json_dict:_ (user : User_model.user) =
+    (* TODO use uuid in the future *)
+    user_unikernel albatross ~user_name:user.name ~unikernel_name:name
+    >>= fun unikernel_info ->
+    match unikernel_info with
+    | None ->
+        Middleware.redirect_to_error ~data:(`String "Builder_web request error")
+          ~title:("An error occured while fetching " ^ name)
```
hannesm, line 1053: Neither data nor title sound like I can do something about it.

unikernel.ml
```diff
@@ -973,62 +995,189 @@ struct
...
+          ~api_meth:false `Internal_server_error reqd ()
+    | Some (unikernel_name, unikernel) -> (
+        Builder_web.send_request http_client
+          (Builder_web.base_url ^ "/hash?sha256=" ^ Ohex.encode unikernel.digest)
+        >>= function
+        | Error (`Msg err) ->
+            Logs.err (fun m -> m "Builder_web update error %s" err);
+            Middleware.redirect_to_error
+              ~data:(`String ("Builder_web request: " ^ err))
+              ~title:(Vmm_core.Name.to_string unikernel_name ^ " update Error")
```
hannesm, line 1063: again, an error I don't understand.

unikernel.ml
```diff
@@ -973,62 +995,189 @@ struct
...
+        | Ok response_body -> (
+            match
+              Builder_web.build_of_json (Yojson.Basic.from_string response_body)
+            with
+            | Error (`Msg err) ->
+                Logs.err (fun m -> m "Builder_web update error %s" err);
+                Middleware.redirect_to_error
+                  ~data:(`String ("Builder_web request: " ^ err))
+                  ~title:
+                    (Vmm_core.Name.to_string unikernel_name ^ " update Error")
```
hannesm, line 1074: same as above. Maybe give some hints what failed. After all, having distinct error messages in distinct code paths helps us developers to figure out where the issue is.

unikernel.ml
```diff
@@ -973,62 +995,189 @@ struct
...
+                  (Builder_web.base_url ^ "/job/" ^ current_job_data.job
+                 ^ "/build/latest")
+                >>= function
+                | Error (`Msg err) ->
+                    Logs.err (fun m -> m "Builder_web update error %s" err);
+                    Middleware.redirect_to_error
+                      ~data:(`String ("Builder_web request: " ^ err))
+                      ~title:
+                        (Vmm_core.Name.to_string unikernel_name
+                        ^ " update Error")
```
hannesm, line 1087: see above

unikernel.ml
```diff
@@ -973,62 +995,189 @@ struct
...
+                      Builder_web.build_of_json
+                        (Yojson.Basic.from_string response_body)
+                    with
+                    | Error (`Msg err) ->
+                        Logs.err (fun m -> m "Builder_web update error %s" err);
+                        Middleware.redirect_to_error
+                          ~data:(`String ("Builder_web request: " ^ err))
+                          ~title:
+                            (Vmm_core.Name.to_string unikernel_name
+                            ^ "update Error")
```
hannesm, line 1100: see above

unikernel.ml
```diff
@@ -973,62 +995,189 @@ struct
...
+                           ^ "")
+                          >>= function
+                          | Error (`Msg err) ->
+                              Logs.err (fun m ->
+                                  m "Builder_web update error %s" err);
+                              Middleware.redirect_to_error
+                                ~data:(`String ("Builder_web request: " ^ err))
+                                ~title:
+                                  (Vmm_core.Name.to_string unikernel_name
+                                  ^ " update Error")
```
hannesm, line 1132: see above

unikernel.ml
```diff
@@ -973,62 +995,189 @@ struct
...
+                                           ~header_list:
+                                             [ ("X-MOLLY-CSRF", csrf) ]
+                                           `OK)
+                                  | Error err ->
+                                      Lwt.return
+                                        (reply reqd ~content_type:"text/html"
+                                           (Guest_layout.guest_layout
+                                              ~page_title:"500 | Mollymawk"
+                                              ~content:
+                                                (Error_page.error_layout err)
```
hannesm, line 1167: and here suddenly the error is exposed to the user. does it make sense, or does it need more context?

unikernel.ml
```diff
@@ -973,62 +995,189 @@ struct
...
+                                           `Internal_server_error))
+                              | Error (`Msg err) ->
+                                  Logs.err (fun m ->
+                                      m "Builder_web update error %s" err);
+                                  Middleware.redirect_to_error
+                                    ~data:
+                                      (`String ("Builder_web request: " ^ err))
+                                    ~title:
+                                      (Vmm_core.Name.to_string unikernel_name
+                                      ^ " update Error")
```
hannesm, line 1178: see above

builder_web.ml
```diff
@@ -0,0 +1,331 @@
...
+
+let ( let* ) = Result.bind
+let base_url = "https://builds.robur.coop"
+
+type build = {
+  job : string;
+  uuid : string;
+  platform : string;
+  start_time : Ptime.t option;
+  finish_time : Ptime.t option;
```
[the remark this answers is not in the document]
PizieDust: We use `Ptime.of_rfc3330` to convert from `strings` to `Ptime.t` which returns a result type.

unikernel.ml
PizieDust: The error here is comes from generating the csrf token and includes the infromation

PizieDust:
The latest commit will display albatross error messages:

image

Is there some sensitive information albatross could emit in errors that we may not want to show the user?

unikernel_single.ml
PizieDust: I resulted to just passing the name I needed to unikernel single instead. Does this work or will more refactoring be good? I am happy to rework the workflow on mollymawk if this is the case :)

hannesm:
> @hannesm This [commit](https://github.com/robur-coop/mollymawk/pull/102/commits/0c973bee8792ef3d54e00fd14a9b51900a59a259) will display albatross error messages:
> Is there some sensitive information albatross could emit in errors that we may not want to show the user?

Sorry, I'm slightly confused - it may help to have some more context:
- what is the request?
- why does it fail?

Now, considering the request being about `roburtls` -- the user will likely be confused to see `robur:roburtls` which they have never put anywhere.

builder_web.ml
```diff
@@ -0,0 +1,331 @@
...
+      | _ ->
+          Error
+            (`Msg
+              ("invalid json for builder_web diff: "
+              ^ Yojson.Basic.to_string (`Assoc xs))))
+  | js ->
+      Error
+        (`Msg
+          ("invalid json for builder_web diff, expected a dict: "
+         ^ Yojson.Basic.to_string js))
```
hannesm, line 316: I didn't look into the json decoding at all.

builder_web.ml
```diff
@@ -0,0 +1,331 @@
...
+          Some (`String uuid),
+          Some (`String platform),
+          Some (`String start_time_str),
+          Some (`String finish_time_str),
+          Some (`Bool main_binary),
+          main_binary_size ) ->
+          let start_time =
+            match Utils.TimeHelper.ptime_of_string start_time_str with
+            | Ok ptime -> Some ptime
+            | Error _ -> None
```
hannesm, line 204: we slurp the error here -- shouldn't we instead fail with the entire decoding (see comment above)

builder_web.ml
```diff
@@ -0,0 +1,331 @@
...
+            (`Msg
+              ("invalid json for builder_web diff: "
+              ^ Yojson.Basic.to_string (`Assoc xs))))
+  | js ->
+      Error
+        (`Msg
+          ("invalid json for builder_web diff, expected a dict: "
+         ^ Yojson.Basic.to_string js))
+
+let send_request http_client url =
```
hannesm, line 318: should we use the `base_url` here directly? i.e. in the code below have a `let url = base_url ^ path in` where `path` is the argument here -- this would ensure that only base_url is accessed, and user code doesn't need to refer to Builder_web.base_url manually.

builder_web.ml
```diff
@@ -0,0 +1,331 @@
...
+    ~headers:[ ("Accept", "application/json") ]
+    url body_f body
+  >>= function
+  | Error (`Msg err) -> Lwt.return (Error (`Msg err))
+  | Error `Cycle -> Lwt.return (Error (`Msg "returned cycle"))
+  | Error `Not_found -> Lwt.return (Error (`Msg "returned not found"))
+  | Ok (resp, body) -> (
+      match resp.Http_mirage_client.status with
+      | `OK -> Lwt.return (Ok body)
+      | _ -> Lwt.return (Error (`Msg resp.reason)))
```
hannesm, line 331: is the reason sufficient? should the status be transferred as well?

middleware.ml
```diff
@@ -32,16 +32,14 @@ let apply_middleware middlewares handler =
...
     let session_header =
       if clear_session then
         [
-          ( "Set-Cookie",
-            User_model.session_cookie
-            ^ "=; Path=/; HttpOnly=true; Expires=2023-10-27T11:00:00.778Z" );
+          ("Set-Cookie", User_model.session_cookie ^ "=; Path=/; HttpOnly=true;");
         ]
       else []
     in
```
hannesm, line 45: I don't understand the above changes, and why they're part of this PR. but it's fine to keep them.

unikernel.ml
```diff
@@ -301,6 +304,39 @@ struct
...
+        (None, Some message)
+    | Ok reply ->
+        let message =
+          Printf.sprintf
+            "Trying to fetch %s: expected a unikernel info reply, received %s"
+            unikernel_name
+            (Format.asprintf "%a" (Vmm_commands.pp_wire ~verbose:false) reply)
+        in
+        Logs.err (fun m -> m "%s" message);
+
```
hannesm, line 337: why is here a newline?

builder_web.ml
PizieDust: With `http_lwt_client` there is a `pp_` function which prints the entire response. I was wondering if we can update `http_mirage_client` to do the same. It lacks a `pp_` function for responses.


But yet, we can also pass the status

unikernel.ml
```diff
@@ -301,6 +304,39 @@ struct
               reply);
         []
 
+  let user_unikernel albatross ~user_name ~unikernel_name =
```
hannesm, line 307: the return value is `(Name.t * Unikernel.info) option * string option` -- where only ever exactly one option is set to `Some`.

We could use a `(Name.t * Unikernel.info, string) result` instead -- and the `Ok` and `Error ` constructors.

middleware.ml
PizieDust: I observed that some cookies sometimes caused weird issues on the browser, so I removed the Expires attribute so the cookies just expire with the browser session which is the default behavior anyways

unikernel.ml
```diff
@@ -973,62 +1009,254 @@ struct
...
+    | _ ->
+        Logs.err (fun m ->
+            m "n unknown error occured while fetching %s from albatross." name);
+        Middleware.redirect_to_error
+          ~data:
+            (`String
+              "An unknown error occured while fetching this unikernel from \
+               albatross. ")
+          ~title:(name ^ " update Error") ~api_meth:false `Internal_server_error
+          reqd ()
```
hannesm, line 1068: if we use a result type, as suggested above, we can remove this code block.

unikernel.ml
```diff
@@ -973,62 +1009,254 @@ struct
...
+        Middleware.redirect_to_error
+          ~data:
+            (`String
+              ("An error occured while fetching " ^ name
+             ^ " from albatross with error " ^ err))
+          ~title:"Albatross Error" ~api_meth:false `Internal_server_error reqd
+          ()
+    | Some (unikernel_name, unikernel), None -> (
+        Builder_web.send_request http_client
+          (Builder_web.base_url ^ "/hash?sha256=" ^ Ohex.encode unikernel.digest)
```
hannesm, line 1086: as suggested in `Builder_web`, we can drop the `Builer_web.base_url` here and always prefix in `send_request`.

unikernel.ml
```diff
@@ -973,62 +1009,254 @@ struct
...
+    | _ ->
+        Logs.err (fun m ->
+            m "n unknown error occured while fetching %s from albatross." name);
+        Middleware.redirect_to_error
+          ~data:
+            (`String
+              "An unknown error occured while fetching this unikernel from \
+               albatross. ")
+          ~title:(name ^ " update Error") ~api_meth:false `Internal_server_error
+          reqd ()
```
hannesm, line 1259: same as above, this case is unnecessary as far as I can see.

unikernel.ml
```diff
@@ -973,62 +1009,254 @@ struct
...
+            Logs.err (fun m ->
+                m
+                  "Builder_web: Error while fetching the current build info of \
+                   %s with error: %s"
+                  name err);
+            Middleware.redirect_to_error
+              ~data:
+                (`String
+                  ("An error occured while fetching the current build \
+                    information from Builder_web. The error gotten is: " ^ err))
```
hannesm, line 1098: I'd drop the `gotten` -- but I'm not a native English speaker.

unikernel.ml
```diff
@@ -973,62 +1009,254 @@ struct
...
+                                     diff between the current and latest build \
+                                     info of %s with error: %s"
+                                    name err);
+                              Middleware.redirect_to_error
+                                ~data:
+                                  (`String
+                                    ("An error occured while fetching the diff \
+                                      between the latest and the current build \
+                                      information from Builder_web. The error \
+                                      gotten is: " ^ err))
```
hannesm, line 1181: as above -- I don't think that a user of mollymawk understands what `Builder_web` is -- but providing the URL may give some insights.

PizieDust:
@hannesm please feel free to merge at anytime :)

hannesm:
merging. we will revise in subsequent PRs.

PR #103, opened 2025-01-19
Use setup-ocamlv3 in ocamlformat workflow

PizieDust:
- This updates our ocamlformat workflow to use `setup-ocamlv3`
- We have some issues with the `darcs` package which causes the ci workflow to fail.  Specifically, the error is


PizieDust:
I'll go ahead and merge as it's a little change. :)

PR #104, opened 2025-01-21
in builder_web: ptime.t is not optional, report error if the json didn't contain a valid ptime.t

hannesm:
update to ocamlformat 0.27.0

/cc @PizieDust


PizieDust:
Thank you @hannesm 
I will go ahead and merge this.

PR #105, opened 2025-01-29
Update unikernel with new build from builds.robur.coop

PizieDust:
This PR is a continuation of the work started in https://github.com/robur-coop/mollymawk/pull/102 and is related to #54 

In this PR, we :
- [x] download the binary from builds.robur.coop,
- [x] perform a manifest check to ensure that there's no changes in the bridge or block devices.
- [x] check that a binary exists from the latest build before attempting to download it
- [ ] check if the vm should be compressed or not
- [x] modify or update the `args` passed in the config to albatross
if everything checks out, then we pass this binary to albatross to perform the update.


unikernel.ml
```diff
@@ -1240,6 +1241,112 @@ struct
...
+                  _;
+                } ) -> (
+            Builder_web.send_request http_client
+              ("/job/" ^ job ^ "/build/" ^ build ^ "/main-binary")
+            >>= function
+            | Error (`Msg err) ->
+                Logs.err (fun m ->
+                    m
+                      "builds.robur.coop: Error while fetching the binary of \
+                       with error: %s"
```
reynir, line 1285: This error message is missing some words it seems: "the binary of with error:"

unikernel.ml
```diff
@@ -1240,6 +1241,112 @@ struct
...
+                      err);
+                Middleware.http_response reqd ~title:"Error"
+                  ~data:
+                    (`String
+                       ("An error occured while fetching the binary from \
+                         builds.robur.coop with error " ^ err))
+                  `Internal_server_error
+            | Ok image -> (
+                Logs.info (fun m -> m "The binary is %s" image);
```
reynir, line 1294: I think this would log a binary image!! Too big :D

albatross.ml
```diff
@@ -91,6 +95,38 @@ struct
...
+    | [], blocks ->
+        Error (`Msg ("Block devices mismatch: " ^ String.concat "," blocks))
+    | bridges, [] ->
+        Error (`Msg ("Network devices mismatch: " ^ String.concat "," bridges))
+    | bridges, blocks ->
+        Error
+          (`Msg
+             ("Block devices mismatch: " ^ String.concat "," blocks
+            ^ " and network devices mismatch: " ^ String.concat "," bridges))
+
```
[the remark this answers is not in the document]
reynir: So you could check the set difference both ways is empty:

```OCaml
let req_only_bridges = String_set.diff bridges bridges'
and mft_only_bridges = String_set.diff bridges' bridges in
...
match (req_only_bridges, mft_only_bridges, req_only_blocks, mft_only_blocls) with
| [], [], [], [] -> Ok ()
| req_only_bridges, mft_only_bridges, ... -> Error (`Msg "Network devices only in manifest: "^String.concat mft_only_bridges ^"...
```

reynir (changes requested):

unikernel.ml
PizieDust: This slipped into the commit by error. Will remove it. Thank you for catching it.

albatross.ml
```diff
@@ -97,35 +97,91 @@ struct
...
+             ^ " and block devices only in manifest: "
+             ^ String.concat "," mft_only_blocks))
+    | req_only_bridges, [], mft_only_blocks, [] ->
+        Error
+          (`Msg
+             ("Network devices missing from manifest: "
+             ^ String.concat "," req_only_bridges
+             ^ " and block devices only in manifest: "
+             ^ String.concat "," mft_only_blocks))
+    | req_only_bridges, req_only_blocks, mft_only_bridges, mft_only_blocks ->
```
reynir, line 174: I think this could be refactored (e.g. "_ devices missing from manifest", "_ devices only in manifest"), that is, factor out the device type.

reynir (approved):
I think this looks fine. I have not tried to review the javascript changes, and I may be missing context.

albatross.ml
```diff
@@ -91,6 +95,94 @@ struct
             }
     | Error (`Msg err) -> Error err
 
+  let manifest_devices_match ~bridges ~block_devices binary =
+    let* mft : Solo5_elftool.mft = Solo5_elftool.query_manifest binary in
```
reynir, line 99: Here we (unfortunately) need to convert the binary to a `Owee_buf.t` (bigarray):
```suggestion
  let manifest_devices_match ~bridges ~block_devices binary =
    let b = Bigarray.Array1.create Bigarray.Int8_unsigned Bigarray.c_layout (String.length b) in
    for i = 0 to String.length b - 1 do
      buf.{i} <- String.get_uint8 b i
    done;
    let binary = b in
    let* mft : Solo5_elftool.mft = Solo5_elftool.query_manifest binary in
```
Otherwise we get a type error. With this PR we should be able to do it without copying too much: https://git.robur.coop/robur/ocaml-solo5-elftool/pulls/2
So while this is very inefficient I'll submit a fix to that Soon™.

reynir (changes requested):

reynir:
Sorry there were still some C stubs in the owee fork that broke things. It's fixed in https://github.com/robur-coop/owee/commit/b1f94753a6a873a21103cb0d19a501b77cdf28d2

reynir (approved):

albatross.ml
PizieDust: Yes I'll send in a commit with better/shorter messages

PizieDust:
@hannesm Please feel free to merge :)

hannesm:
Thanks for your reminder, @PizieDust. I didn't look at this change at all. If you'd like to have this PR reviewed by me, it will need to wait some more time.

OTOH, Reynir reviewed it, so it should be good. I'd appreciate if we avoid a pin in `config.ml` and wait for the solo5-elftool 0.4 to be released (which, if I understand Reynir, will happen soon)!?

PizieDust:
> Thanks for your reminder, @PizieDust. I didn't look at this change at all. If you'd like to have this PR reviewed by me, it will need to wait some more time.
> OTOH, Reynir reviewed it, so it should be good. I'd appreciate if we avoid a pin in `config.ml` and wait for the solo5-elftool 0.4 to be released (which, if I understand Reynir, will happen soon)!?

Yes we shouldn't merge with the pin. We'll wait for the release.

albatross.ml
```diff
@@ -91,6 +95,98 @@ struct
...
 
+  let manifest_devices_match ~bridges ~block_devices binary =
+    let b =
+      Bigarray.Array1.create Bigarray.Int8_unsigned Bigarray.c_layout
+        (String.length binary)
+    in
+    for i = 0 to String.length binary - 1 do
+      b.{i} <- String.get_uint8 binary i
+    done;
+    let binary = b in
```
hannesm, line 106: can we name the binding `b` above just `binary` and then don't need this alias?

reynir:
This will need more changes for ocaml-solo5-elftool then. I'll work on that
and then merge (I didn't yet look at the new error message code).

tirs. 4. feb. 2025 16.46 skrev Hannes Mehnert ***@***.***>:

albatross.ml
```diff
@@ -91,6 +95,98 @@ struct
...
+             ^ String.concat "," mft_only_blocks))
+    | req_only_bridges, req_only_blocks, mft_only_bridges, mft_only_blocks ->
+        Error
+          (`Msg
+             (String.concat "," req_only_bridges
+             ^ String.concat "," req_only_blocks
+             ^ " devices missing from manifest and "
+             ^ String.concat "," mft_only_bridges
+             ^ String.concat "," mft_only_blocks
+             ^ " devices only in manifest."))
```
hannesm, line 188: you refer to `manifest` in this error message a lot of times, but is this really the wording an operator of mollymawk will understand?

As I understand it, we have the unikernel image and in there we have required devices (what you name manifest); and then we have the command line parameters. So these two sets (actually 4, since it is two for bridges and two for block devices) may disjoint, and if they are the unikernel won't start.

What would be a message an operator understands? Maybe
"You didn't specify the network bridge 'management', which is required by the unikernel."

And
"You specified the network bridge 'private' as command line parameter, which the unikernel does not refer to at all."

Now, it may help to suggest what the operator needs to do:
"Please add a command-line parameter for the 'management' bridge: '--net:private=<my-bridge>'".

"Please remove the command-line parameter '--net=private'"

albatross_json.ml
```diff
@@ -298,3 +298,63 @@ let config_of_json str =
...
+       ("memory", `Int unikernel.memory);
+       ("network_interfaces", `List (List.map bridge_to_json unikernel.bridges));
+       ( "block_devices",
+         `List (List.map block_device_to_json unikernel.block_devices) );
+     ]
+    @
+    match unikernel.argv with
+    | Some args ->
+        [ ("arguments", `List (List.map (fun arg -> `String arg) args)) ]
+    | None -> [])
```
hannesm, line 360: there's in Albatross_json right at the top (line 3) a `unikernel_info` function -- how are they related?

unikernel.ml
```diff
@@ -1240,6 +1241,122 @@ struct
...
+                 builds.robur.coop with error " ^ err))
+          `Internal_server_error
+    | Ok image -> (
+        match
+          Albatross.manifest_devices_match ~bridges:unikernel_cfg.bridges
+            ~block_devices:unikernel_cfg.block_devices image
+        with
+        | Error (`Msg err) ->
+            Middleware.http_response reqd ~title:"Error"
+              ~data:(`String ("Manifest mismatch: " ^ err))
```
hannesm, line 1269: as outlined above, I wouldn't introduce the `Manifest` teminology.

unikernel.ml
```diff
@@ -1240,6 +1241,122 @@ struct
...
+        | Ok () -> (
+            let unikernel_config = { unikernel_cfg with image } in
+            Albatross.query albatross ~domain:user.name ~name:unikernel_name
+              (`Unikernel_cmd (`Unikernel_force_create unikernel_config))
+            >>= function
+            | Error msg ->
+                Logs.err (fun m -> m "Error querying albatross: %s" msg);
+                Middleware.http_response reqd ~title:"Error"
+                  ~data:(`String ("Error querying albatross: " ^ msg))
+                  `Internal_server_error
```
hannesm, line 1280: this error could be a bit more precise -- we're 'force_create' a unikernel here.

unikernel.ml
```diff
@@ -1240,6 +1241,122 @@ struct
...
+                | Ok res ->
+                    Logs.info (fun m ->
+                        m "%s has been updated succesfully with result: %s"
+                          unikernel_name
+                          (Yojson.Basic.to_string res));
+                    Middleware.http_response reqd ~title:"Update Successful"
+                      ~data:
+                        (`String
+                           (unikernel_name
+                          ^ " has been updated to the latest build."))
```
hannesm, line 1296: maybe to build <ID> where ID could be the sha256 or the build UUID so that the operator could look it up (since there may be while updating an even newer build be produced and uploaded to builds.robur.coop).

unikernel.ml
```diff
@@ -1240,6 +1241,122 @@ struct
...
+                | Ok cfg ->
+                    process_unikernel_update ~unikernel_name ~job ~build cfg
+                      user albatross http_client reqd
+                | Error (`Msg err) ->
+                    Logs.warn (fun m -> m "Couldn't decode data %s" err);
+                    Middleware.http_response reqd ~title:"Error"
+                      ~data:(`String (String.escaped err))
+                      `Internal_server_error))
+        | Ok (Some args) -> (
+            match Albatross_json.config_of_json args with
```
hannesm, line 1346: I don't quite understand the argument handling:
```OCaml
match Utils.Json.get "unikernel_arguments" json_dict with
    |  arguments -> match Utils.Json.string_or_none "unikernel_arguments" arguments with
      | OK (Some args) -> match Albatross_json.config_of_json args with
```

so in which way need they to be present? json encoded that is encoded as string? why not just as a dict directly in the user parameters?

reynir:
The code is now updated to ocaml-solo5-elftool.0.4.0 (pending release). With a temporary pin (now removed) it passes CI \o/

PizieDust:
> The code is now updated to ocaml-solo5-elftool.0.4.0 (pending release). With a temporary pin (now removed) it passes CI \o/

Awesome. 
Thank you Reynir

albatross.ml
PizieDust: I totally agree with this. I have a question, for a missing bridge or block device, does the user add it in the block_devices/network_devices section this json, or in the `arguments`. I am asking based on the example you have given which uses command-line parameters. At the moment the only way the user can modify the arguments is by modifying a json such as below.

```json
{
  "name": "some name",
  "typ": "solo5",
  "fail_behaviour": {
    "restart": null,
    "exit_code": [],
    "all_exit_codes": true
  },
  "cpuid": 16,
  "memory": 1024,
  "block_devices": [
    { "name": "data", "host_device": "service", "sector_size": 512 }
  ],
  "network_interfaces": [
    {
      "name": "service",
      "host_device": "internal",
    }
  ],
  "arguments": [
    "--ipv4=192.168.20.11/00", 
    "--ignore-local-git"
  ],
}

```

reynir: the `arguments` field is the boot arguments to the unikernel. The missing bridge would go into the `network_interfaces` list. Similarly, missing block devices go into the `block_devices` list.

unikernel.ml
PizieDust: This is done in the current commit

hannesm:
This is fine. I don't quite understand why it triggers the opam-monorepo failure, though.

I'll merge to clean up the queue, and will propose the step for albatross 2.4.0

PR #106, opened 2025-02-10
update to albatross 2.4.0 changes

hannesm:
cc @PizieDust

### DIFF
--- a/account_usage.ml
+++ b/account_usage.ml
@@ -47,15 +47,14 @@ let account_usage_layout policy unikernels blocks =
     List.iter
       (fun (_, unikernel) ->
         List.iter
-          (fun (name, bridge, _) ->
-            let bridge_name = Option.value bridge ~default:name in
+          (fun { Vmm_core.Unikernel.host_device; _ } ->
             (* Only count if the bridge is in policy.bridges *)
-            if Vmm_core.String_set.mem bridge_name policy.bridges then
+            if Vmm_core.String_set.mem host_device policy.bridges then
               let count =
-                Hashtbl.find_opt bridge_count bridge_name
+                Hashtbl.find_opt bridge_count host_device
                 |> Option.value ~default:0
               in
-              Hashtbl.replace bridge_count bridge_name (count + 1))
+              Hashtbl.replace bridge_count host_device (count + 1))
           unikernel.Vmm_core.Unikernel.bridges)
       unikernels;
 
--- a/albatross_json.ml
+++ b/albatross_json.ml
@@ -16,25 +16,27 @@ let unikernel_info (unikernel_name, info) =
   and cpuid c = `Int c
   and memory m = `Int m
   and block_devices bs =
-    let block (name, dev, sec) =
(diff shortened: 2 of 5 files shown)


PizieDust:
I have some package conflict issues while upgrading to `albatross 2.4.0`

`solo5-elftool.0.4.0` requires `cachet 0.0.1` while `albatross 2.4.0` requires `cachet 0.0.2`

hannesm:
The `cachet` issue shouldn't be any issue.

The `ptime` issue, see the links provided above. As interim solution `opam pin add albatross.2.4.0 --dev` - where the opam file is fixed (to be fine with ptime 1.1.0).

hannesm:
CI should be fine now that https://github.com/ocaml/opam-repository/pull/27421 got merged

PizieDust:
> CI should be fine now that [ocaml/opam-repository#27421](https://github.com/ocaml/opam-repository/pull/27421) got merged

Yes thank you, 
I added some minor changes for the old unikernel info format.

hannesm:
Thanks @PizieDust - I suspect we'll need to support both the old_unikernel_info and the new unikernel_info... or only the new (depending on where we expect mollymawk to be compatible with -- on our server we already have the albatross 2.4 which will use the new unikernel_infi).

in the long run we need to decide on a cut for albatross to collect all the old no longer used (or unlikely used) wire formats.

I'm personally fine if we decide for mollymawk to only work with the most recent release of albatross - and error out if we receive old replies. what do you think? what motivated your change?

PizieDust:
> Thanks @PizieDust - I suspect we'll need to support both the old_unikernel_info and the new unikernel_info... or only the new (depending on where we expect mollymawk to be compatible with -- on our server we already have the albatross 2.4 which will use the new unikernel_infi).
> 
> in the long run we need to decide on a cut for albatross to collect all the old no longer used (or unlikely used) wire formats.
> 
> I'm personally fine if we decide for mollymawk to only work with the most recent release of albatross - and error out if we receive old replies. what do you think? what motivated your change?

Yes I think this makes sense. We should only support the latest release of albatross.
However, the latest replies for `Unikernel_info` are not compatible with mollymawk and I think the update should come from albatross. This is why I temporarily reverted to `Old unikernel info`

PizieDust:
With the latest `Unkernel_info`, we get this error on Albatross:

```
2025-02-12T21:37:46+01:00: [ERROR] [application] expected a unikernel info reply, received host [unikernel: robur:]: [unikernel: :roburtls]: typ solo5
```

which comes from https://github.com/robur-coop/mollymawk/blob/main/unikernel.ml#L299-L305

hannesm:
oh why are they not compatible? what is the compile error?

I pushed to main a fix to cirrus.yml, could you merge or rebase this PR onto main (I don't seem to be able from my mobile phone to do so). thanks :)

PizieDust (approved):
Thank you @hannesm

PizieDust:
We can update to the new `Unikernel_info` reply in a subsequent PR :)

PR #110, opened 2025-02-18
Use accept: application/json

reynir:
When doing requests in the frontend to the server specify the accept header with appropriate types. For example, if we expect json let the server know. This allows the server to generate errors that are json objects.


PizieDust (approved):
Thank you @reynir

PR #111, opened 2025-02-25
Rollback functionality of Unikernels

PizieDust:
This PR implements Rollback functionality for unikernel updates.
The methodology is that, if we try to update a unikernel, we store some metadata about the running unikernel to disk and then attempt the update. If this fails, we call the rollback function with the previous running build.

### DIFF
--- a/assets/main.js
+++ b/assets/main.js
@@ -982,6 +982,45 @@ async function updateUnikernel(job, latest_build, current_build, unikernel_name)
 	}
 }
 
+async function rollbackUnikernel(unikernel_name) {
+	const rollbackButton = document.getElementById("unikernel-rollback");
+	const molly_csrf = document.getElementById("molly-csrf").value;
+	const formAlert = document.getElementById("form-alert");
+	try {
+		buttonLoading(rollbackButton, true, "Rolling back...")
+		const response = await fetch("/api/unikernel/rollback", {
+			method: 'POST',
+			headers: {
+				"Content-Type": "application/json",
+				"Accept": "application/json",
+			},
+			body: JSON.stringify(
+				{
+					"unikernel_name": unikernel_name,
+					"molly_csrf": molly_csrf
+				})
+		})
+		const data = await response.json();
+		if (data.status === 200) {
+			postAlert("bg-primary-300", "Unikernel rollback succesful");
+			setTimeout(() => window.location.reload(), 1000);
+			buttonLoading(rollbackButton, false, "Rollback")
+		} else {
+			postAlert("bg-secondary-300", data.data);
+			formAlert.classList.remove("hidden", "text-primary-500");
+			formAlert.classList.add("text-secondary-500");
+			formAlert.textContent = data.data
+			buttonLoading(rollbackButton, false, "Rollback")
+		}
+	} catch (error) {
+		postAlert("bg-secondary-300", error);
+		formAlert.classList.remove("hidden", "text-primary-500");
+		formAlert.classList.add("text-secondary-500");
+		formAlert.textContent = data.data
+		buttonLoading(rollbackButton, false, "Rollback")
+	}
+}
+
 function isValidName(s) {
 	const length = s.length;
 	if (length === 0 || length >= 64) return false;
--- a/assets/main.js
+++ b/assets/main.js
@@ -932,7 +932,7 @@ async function updateToken(value) {
 	}
 }
 
-async function updateUnikernel(job, build, unikernel_name) {
+async function updateUnikernel(job, latest_build, current_build, unikernel_name) {
 	const updateButton = document.getElementById("update-unikernel-button");
 	const unikernelArguments = document.getElementById("unikernel-arguments").value;
 	const argumentsToggle = document.getElementById("arguments-toggle").checked;
@@ -954,7 +954,8 @@ async function updateUnikernel(job, build, unikernel_name) {
 			body: JSON.stringify(
 				{
 					"job": job,
-					"build": build,
+					"latest_build": latest_build,
+					"current_build": current_build,
 					"unikernel_name": unikernel_name,
 					"unikernel_arguments": argumentsToggle ? JSON.parse(unikernelArguments) : null,
 					"molly_csrf": molly_csrf
--- a/storage.ml
+++ b/storage.ml
@@ -1,6 +1,6 @@
 open Utils.Json
 
-let current_version = 6
+let current_version = 7
 (* version history:
    1 was initial (fields until email_verification_uuid)
    2 added active
@@ -8,6 +8,7 @@ let current_version = 6
    4 properly serialised super_user
    5 cookie has two new fields last_access and user_agent
    6 tokens has 3 new fields: name, last_access and usage_count
+   7 added unikernel_updates to keep track of when unikernels are updated
 *)
 
 let t_to_json users configuration =
@@ -26,6 +27,7 @@ let t_of_json json =
       | Some (`Int v), Some (`List users), Some configuration ->
           let* () =
             if v = current_version then Ok ()
+            else if v = 6 then Ok ()
             else if v = 5 then Ok ()
             else if v = 4 then Ok ()
             else if v = 3 then Ok ()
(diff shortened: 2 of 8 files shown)


hannesm:
Great to see this PR.

I struggle with the naming "current_build"and "latest_build". Can we have some more descriptive terms?

Why should we decode and encode the config to/from json, if albatross already provides asn1 decoders and encoders for that data?

You now store a list of "unikernel_updates" in the user configuration. Is this list ever getting smaller? I'm not sure what we did for cookies, but I'm a bit scared of ever-growing lists that are stored to persistent storage.

Semantics wise: I'm not sure what is expected to happen when you update twice? I guess I don't understand when a rollback is triggered, and whether in the list of updates stored persistently a unikernel identified by its name can be present twice and if yes why.

PizieDust:
Thank you

> Great to see this PR.
> 
> I struggle with the naming "current_build"and "latest_build". Can we have some more descriptive terms?
> 

Yes we can change the names But just to explain a bit, by `current_build` I mean the version of the unikernel currently running and by `latest_build` I mean the version that we want to upgrade to.

> Why should we decode and encode the config to/from json, if albatross already provides asn1 decoders and encoders for that data?
> 
> You now store a list of "unikernel_updates" in the user configuration. Is this list ever getting smaller? I'm not sure what we did for cookies, but I'm a bit scared of ever-growing lists that are stored to persistent storage.
> 
> Semantics wise: I'm not sure what is expected to happen when you update twice? I guess I don't understand when a rollback is triggered, and whether in the list of updates stored persistently a unikernel identified by its name can be present twice and if yes why.

For this list, only one instance of a unikernel can be stored. so only one instance of a unikernel can exist in the list. If we update twice, the list will contain data related to the latest update only.

storage.ml
```diff
@@ -208,6 +212,23 @@ module Make (BLOCK : Mirage_block.S) = struct
         Logs.err (fun m -> m "Error with storage: %s" err);
         Lwt.return (Error (`Msg err))
 
+  let update_user_unikernel_updates store
+      (new_update : User_model.unikernel_update) (user : User_model.user) =
+    let is_unique (u : User_model.unikernel_update) =
+      u.name <> new_update.name
```
hannesm, line 218: could we please avoid the polymorphic comparison, and be more explicit? `not (String.equal u.name new_update.name)` should be fine here.

unikernel.ml
```diff
@@ -1328,17 +1386,41 @@ struct
...
+                    | Ok _res ->
+                        Middleware.http_response reqd ~title:"Update Successful"
+                          ~data:
+                            (`String
+                               ("Succesfully updated " ^ unikernel_name
+                              ^ "to build " ^ latest_build))
+                          `OK
+                    | Error (`Msg _err, _status) ->
+                        unikernel_rollback ~unikernel_name albatross store
+                          http_client reqd user)
```
hannesm, line 1403: Having this here means that if anything in `process_unikernel_change` fails, we do a rollback. Is this desired?

From reading the code, I have the feeling that `process_unikernel_change` does what it tells: quite a lot, including "updating the store" (which may fail), a failed HTTP request to builds.robur.coop, manifest devices match, *force_create*, Albatross_json.res.

Now, only a failed *force_create* should potentially lead to a rollback, no? Because if there's an issue in updating the store, a failed HTTP request, when the manifest mismatches -- we should present the error to the user -- and since nothing has been done in respect to albatross, we should leave the existing unikernel running. Or am I getting this wrong?

Maybe it makes sense to separate the preconditions for deploying a unikernel from the actual deployment?

PizieDust: Yes you have it correctly. 
I will separate the function into different purpose-focused functions. 

The final rollback should only be triggered if `force_create` fails.

hannesm:
> > I struggle with the naming "current_build"and "latest_build". Can we have some more descriptive terms?
> 
> Yes we can change the names But just to explain a bit, by `current_build` I mean the version of the unikernel currently running and by `latest_build` I mean the version that we want to upgrade to.

What about `currently_running_unikernel` and `to_be_updated_unikernel` (I'm not too happy with these names).

> > Why should we decode and encode the config to/from json, if albatross already provides asn1 decoders and encoders for that data?
> 
> when storing the metadata for a unikernel before updating it, i figured its important to store the config so we can always rollback using the same configurations as before. I'll look more into using the asn1 decoders/encoders and update the pr :)

We can as well do that in a later PR. My main question is whether there's any reason to use json for everyhing.

> > You now store a list of "unikernel_updates" in the user configuration. Is this list ever getting smaller? I'm not sure what we did for cookies, but I'm a bit scared of ever-growing lists that are stored to persistent storage.
> > Semantics wise: I'm not sure what is expected to happen when you update twice? I guess I don't understand when a rollback is triggered, and whether in the list of updates stored persistently a unikernel identified by its name can be present twice and if yes why.
> 
> For this list, only one instance of a unikernel can be stored. so only one instance of a unikernel can exist in the list. If we update twice, the list will contain data related to the latest update only.

Ok. But the list will never be cleared when the update was successful.

> Also a rollback is triggered automatically if an update fails because of an unavoidable error. By this I mean, for instance if the user provides a config which will mismatch the manifest, no rollback will be performed, since the `force-create-unikernel` was not triggered. The rollback will mostly only be triggered in cases were `force-create-unikernel` fails.

I don't find what you say here in the code. I commented on the code where I think that a rollback is triggered even though it shouldn't.

PizieDust:
> What about currently_running_unikernel and to_be_updated_unikernel (I'm not too happy with these names).

These names are fine, much better than what I used :)

> We can as well do that in a later PR. My main question is whether there's any reason to use json for everyhing.

No we don't have to use json. I mainly used it since most of the other processes already rely on json so I wanted to reuse as much as I can

> Ok. But the list will never be cleared when the update was successful.

I will update the PR to clear the list :)

hannesm:
Another thing is that we need to ensure to never rollback while we do a rollback - to avoid (infinite) rollback loops.

PizieDust:
> Hmm, and now I struggle with: rollback being triggered when unikernel_force_create fails. But when does it fail, and why should a rollback then be beneficial?
> 
> Isn't there a lot checking done before the unikernel_force_create, so that it is unlikely to fail? i.e. manifest, we could as well look into resource usage / policies.
> 
> Wasn't the purpose of a rollback that either there's a service not responding or an operator figures shortly after the update that something is fishy? I currently think this "automatic rollback" based on the failure of "unikernel_force_create" is not super-useful.

To be fair, I haven't had a single update fail that required a rollback. So in this case it's correct that rolling back during `unikernel-force-create` is not super useful. 

Indeed all the checks are done before the upgrade is done.

PizieDust:
@hannesm I think it's now ready for another review :)
Let me know if there are more changes that will make the PR better.
- At the moment we have the rollback button next to update which is valid for 10 minutes and then it dissaspears.
- We no longer have any automatic rollbacks
- The pre-processing to perform an update is separated from the update itself.
- We have an endpoint to do rollbacks (which will also check that it's within 10 minutes)

assets/main.js
```diff
@@ -954,7 +954,8 @@ async function updateUnikernel(job, build, unikernel_name) {
 			body: JSON.stringify(
 				{
 					"job": job,
-					"build": build,
+					"latest_build": latest_build,
+					"current_build": current_build,
```
hannesm, line 958: again, the naming... so we have a "current_unikernel" and an "update_candidate" maybe?

unikernel.ml
```diff
@@ -1201,70 +1212,172 @@ struct
...
+        | Ok res ->
+            Logs.info (fun m ->
+                m
+                  "albatross-force-create: succesfully created %s with result \
+                   %s"
+                  unikernel_name
+                  (Yojson.Basic.to_string res));
+            Lwt.return (Ok (Yojson.Basic.to_string res)))
+
+  let process_change ~unikernel_name ~job ~to_be_updated_unikernel
```
hannesm, line 1248: I'm curious about the logic: shouldn't we first fetch the unikernel, check that the device matches -- before pushing that information that we updated the unikernel (and make a rollback possible)?

hannesm: I'm also curious: should we schedule it from being removed from the list after the 10 minutes?

hannesm: But maybe that's for later

PR #116, opened 2025-03-22
Perform Liveliness checks

PizieDust:
This PR adds new functionality for performing liveliness checks with automatic rollback functionality.

## How it works
During an update, the user selects which type of liveliness check they want (HTTP, DNS or TCP) and provides an address. This address is then checked that it returns a positive response and then the update is performed, else a rollback is performed. 

At the moment, HTTP and DNS checks have been implemented.

### DIFF
--- a/assets/main.js
+++ b/assets/main.js
@@ -938,11 +938,54 @@ async function updateUnikernel(job, to_be_updated_unikernel, currently_running_u
 	const argumentsToggle = document.getElementById("arguments-toggle").checked;
 	const molly_csrf = document.getElementById("molly-csrf").value;
 	const formAlert = document.getElementById("unikernel-arguments-alert");
+	const liveliness_toggle = document.getElementById("liveliness-toggle").checked;
+	const http_toggle = document.getElementById("http-toggle").checked;
+	const dns_toggle = document.getElementById("dns-toggle").checked;
+	const tcp_toggle = document.getElementById("tcp-toggle").checked;
+	const tcp_address = document.getElementById("tcp-address").value.trim();
+	const dns_address = document.getElementById("dns-name").value.trim();
+	const http_address = document.getElementById("http-address").value.trim();
+
 	if (argumentsToggle && !unikernelArguments) {
-		postAlert("bg-secondary-300", "You must give arguments for this build else switch 'Update the configuration for this build' off");
+		formAlert.classList.remove("hidden", "text-primary-500");
+		formAlert.classList.add("text-secondary-500");
+		formAlert.textContent = "You must give arguments for this build else switch 'Update the configuration for this build' off"
 		buttonLoading(updateButton, false, "Proceed to update")
 		return;
 	}
+
+	if (liveliness_toggle && !(http_toggle || dns_toggle || tcp_toggle)) {
+		formAlert.classList.remove("hidden", "text-primary-500");
+		formAlert.classList.add("text-secondary-500");
+		formAlert.textContent = "Error: If liveliness is enabled, at least one of HTTP, DNS, or TCP must be activated."
+		buttonLoading(updateButton, false, "Proceed to update")
+		return;
+	}
+
+	if (tcp_toggle && !tcp_address) {
+		formAlert.classList.remove("hidden", "text-primary-500");
+		formAlert.classList.add("text-secondary-500");
+		formAlert.textContent = "Error: TCP is enabled, but no TCP address is provided."
+		buttonLoading(updateButton, false, "Proceed to update")
+		return;
+	}
+
+	if (dns_toggle && !dns_address) {
+		formAlert.classList.remove("hidden", "text-primary-500");
+		formAlert.classList.add("text-secondary-500");
+		formAlert.textContent = "Error: DNS is enabled, but no DNS name is provided."
+		buttonLoading(updateButton, false, "Proceed to update")
+		return;
+	}
+
+	if (http_toggle && !http_address) {
+		formAlert.classList.remove("hidden", "text-primary-500");
+		formAlert.classList.add("text-secondary-500");
+		formAlert.textContent = "Error: HTTP is enabled, but no HTTP address is provided."
+		buttonLoading(updateButton, false, "Proceed to update")
+		return;
+	}
+
 	try {
 		buttonLoading(updateButton, true, "Updating...")
 		const response = await fetch("/api/unikernel/update", {
@@ -958,6 +1001,9 @@ async function updateUnikernel(job, to_be_updated_unikernel, currently_running_u
 					"currently_running_unikernel": currently_running_unikernel,
 					"unikernel_name": unikernel_name,
 					"unikernel_arguments": argumentsToggle ? JSON.parse(unikernelArguments) : null,
+					"http_liveliness_address": http_address ?? null,
+					"tcp_liveliness_address": tcp_address ?? null,
+					"dns_liveliness_address": dns_address ?? null,
 					"molly_csrf": molly_csrf
 				})
 		})
--- a/assets/main.js
+++ b/assets/main.js
@@ -992,11 +992,11 @@ async function updateUnikernel(job, to_be_updated_unikernel, currently_running_u
 					"currently_running_unikernel": currently_running_unikernel,
 					"unikernel_name": unikernel_name,
 					"unikernel_arguments": argumentsToggle ? JSON.parse(unikernelArguments) : null,
-					"http_liveliness_address": http_address ? http_address : null,
-					"dns_liveliness": {
+					"http_liveliness_address": http_toggle ? (http_address ? http_address : null) : null,
+					"dns_liveliness": dns_toggle ? ({
 						"dns_address": dns_address ? dns_address : null,
 						"dns_name": dns_name ? dns_name : null
-					},
+					}) : null,
 					"molly_csrf": molly_csrf
 				})
 		})
--- a/assets/main.js
+++ b/assets/main.js
@@ -992,8 +992,11 @@ async function updateUnikernel(job, to_be_updated_unikernel, currently_running_u
 					"currently_running_unikernel": currently_running_unikernel,
 					"unikernel_name": unikernel_name,
 					"unikernel_arguments": argumentsToggle ? JSON.parse(unikernelArguments) : null,
-					"http_liveliness_address": http_address ?? null,
-					"dns_liveliness": { "dns_address": dns_address ?? null, "dns_name": dns_name ?? null },
+					"http_liveliness_address": http_address ? http_address : null,
+					"dns_liveliness": {
+						"dns_address": dns_address ? dns_address : null,
+						"dns_name": dns_name ? dns_name : null
+					},
 					"molly_csrf": molly_csrf
 				})
 		})
--- a/assets/main.js
+++ b/assets/main.js
@@ -941,7 +941,8 @@ async function updateUnikernel(job, to_be_updated_unikernel, currently_running_u
 	const liveliness_toggle = document.getElementById("liveliness-toggle").checked;
 	const http_toggle = document.getElementById("http-toggle").checked;
 	const dns_toggle = document.getElementById("dns-toggle").checked;
-	const dns_address = document.getElementById("dns-name").value.trim();
+	const dns_address = document.getElementById("dns-address").value.trim();
+	const dns_name = document.getElementById("dns-name").value.trim();
 	const http_address = document.getElementById("http-address").value.trim();
 
 	if (argumentsToggle && !unikernelArguments) {
@@ -960,10 +961,10 @@ async function updateUnikernel(job, to_be_updated_unikernel, currently_running_u
 		return;
 	}
 
-	if (dns_toggle && !dns_address) {
+	if (dns_toggle && !(dns_address && dns_name)) {
 		formAlert.classList.remove("hidden", "text-primary-500");
 		formAlert.classList.add("text-secondary-500");
-		formAlert.textContent = "Error: DNS is enabled, but no DNS name is provided."
+		formAlert.textContent = "Error: DNS is enabled, but DNS name and DNS address is not provided."
 		buttonLoading(updateButton, false, "Proceed to update")
 		return;
 	}
@@ -993,6 +994,7 @@ async function updateUnikernel(job, to_be_updated_unikernel, currently_running_u
 					"unikernel_arguments": argumentsToggle ? JSON.parse(unikernelArguments) : null,
 					"http_liveliness_address": http_address ?? null,
 					"dns_liveliness_address": dns_address ?? null,
+					"dns_liveliness_name": dns_name ?? null,
 					"molly_csrf": molly_csrf
 				})
 		})
--- a/assets/main.js
+++ b/assets/main.js
@@ -932,7 +932,7 @@ async function updateToken(value) {
 	}
 }
 
-async function updateUnikernel(job, latest_build, current_build, unikernel_name) {
+async function updateUnikernel(job, to_be_updated_unikernel, currently_running_unikernel, unikernel_name) {
 	const updateButton = document.getElementById("update-unikernel-button");
 	const unikernelArguments = document.getElementById("unikernel-arguments").value;
 	const argumentsToggle = document.getElementById("arguments-toggle").checked;
@@ -954,8 +954,8 @@ async function updateUnikernel(job, latest_build, current_build, unikernel_name)
 			body: JSON.stringify(
 				{
 					"job": job,
-					"latest_build": latest_build,
-					"current_build": current_build,
+					"to_be_updated_unikernel": to_be_updated_unikernel,
+					"currently_running_unikernel": currently_running_unikernel,
 					"unikernel_name": unikernel_name,
 					"unikernel_arguments": argumentsToggle ? JSON.parse(unikernelArguments) : null,
 					"molly_csrf": molly_csrf
--- a/assets/main.js
+++ b/assets/main.js
@@ -941,8 +941,6 @@ async function updateUnikernel(job, to_be_updated_unikernel, currently_running_u
 	const liveliness_toggle = document.getElementById("liveliness-toggle").checked;
 	const http_toggle = document.getElementById("http-toggle").checked;
 	const dns_toggle = document.getElementById("dns-toggle").checked;
-	const tcp_toggle = document.getElementById("tcp-toggle").checked;
-	const tcp_address = document.getElementById("tcp-address").value.trim();
 	const dns_address = document.getElementById("dns-name").value.trim();
 	const http_address = document.getElementById("http-address").value.trim();
 
@@ -954,18 +952,10 @@ async function updateUnikernel(job, to_be_updated_unikernel, currently_running_u
 		return;
 	}
 
-	if (liveliness_toggle && !(http_toggle || dns_toggle || tcp_toggle)) {
+	if (liveliness_toggle && !(http_toggle || dns_toggle)) {
 		formAlert.classList.remove("hidden", "text-primary-500");
 		formAlert.classList.add("text-secondary-500");
-		formAlert.textContent = "Error: If liveliness is enabled, at least one of HTTP, DNS, or TCP must be activated."
-		buttonLoading(updateButton, false, "Proceed to update")
-		return;
-	}
-
-	if (tcp_toggle && !tcp_address) {
-		formAlert.classList.remove("hidden", "text-primary-500");
-		formAlert.classList.add("text-secondary-500");
-		formAlert.textContent = "Error: TCP is enabled, but no TCP address is provided."
+		formAlert.textContent = "Error: If liveliness is enabled, at least one of HTTP or DNS must be activated."
 		buttonLoading(updateButton, false, "Proceed to update")
 		return;
 	}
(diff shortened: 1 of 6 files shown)


assets/main.js
```diff
@@ -932,17 +932,51 @@ async function updateToken(value) {
 	}
 }
 
-async function updateUnikernel(job, latest_build, current_build, unikernel_name) {
+async function updateUnikernel(job, to_be_updated_unikernel, currently_running_unikernel, unikernel_name) {
```
hannesm, line 935: I'm not sure why these renamings are part of this PR. Makes it rather hard to review, but so be it.

unikernel.ml
```diff
@@ -1045,8 +1045,8 @@ struct
           ~title:"Albatross Error" ~api_meth:false `Internal_server_error reqd
           ()
     | Ok (unikernel_name, unikernel) -> (
-        Builder_web.send_request http_client
-          ("/hash?sha256=" ^ Ohex.encode unikernel.digest)
+        Utils.send_http_request http_client ~base_url:Builder_web.base_url
+          ~path:("/hash?sha256=" ^ Ohex.encode unikernel.digest)
```
hannesm, line 1049: same as above, makes the PR rather hard to review. Maybe next time split this into a separate PR?

unikernel.ml
```diff
@@ -1393,15 +1398,21 @@ struct
...
       Utils.Json.
         ( get "job" json_dict,
-          get "latest_build" json_dict,
-          get "current_build" json_dict,
+          get "to_be_updated_unikernel" json_dict,
+          get "currently_running_unikernel" json_dict,
           get "unikernel_name" json_dict,
+          get "http_liveliness_address" json_dict,
+          get "dns_liveliness_address" json_dict,
+          get "dns_liveliness_name" json_dict,
```
hannesm, line 1406: I don't quite understand why the dns_liveliness_address and _name are two separate fields -- the idea is both are either present or they are absent -- why not use one dict entry then (dns_liveliness), and have there two fields?

unikernel.ml
```diff
@@ -1411,87 +1422,140 @@ struct
                 (`String ("Could not get the unikernel arguments json: " ^ err))
               `Bad_request
         | Ok None -> (
```
hannesm, line 1424: just to be sure, we're in the case that an update is being done -- and now we first dispatch whether the "unikernel_arguments" could be a "config" (using the "config_or_none")?

Why would this be done before the Liveliness_check -- which then needs to be called at two sides in this function?

unikernel.ml
```diff
@@ -1411,87 +1422,140 @@ struct
...
-                  Albatross_json.(
-                    unikernel_info (n, unikernel)
-                    |> Yojson.Basic.to_string |> config_of_json)
-                with
-                | Ok cfg -> (
-                    process_change ~unikernel_name ~job ~to_be_updated_unikernel
-                      ~currently_running_unikernel cfg user store http_client
-                      `Update
+                       ("Liveliness check of currentyly running unikernel \
+                         failed with error(s): " ^ err))
```
hannesm, line 1494: why "(s)"?

unikernel.ml
```diff
@@ -1411,87 +1422,140 @@ struct
...
+                  `Bad_request)
+        | Ok (Some cfg) -> (
+            Liveliness_checks.liveliness_checks ~http_liveliness_address
+              ~dns_liveliness_address ~dns_liveliness_name http_client
```
hannesm, line 1498: here's the second call to liveliness_checks of the currently running unikernel.

unikernel.ml
```diff
@@ -1411,87 +1422,140 @@ struct
...
+                      "Liveliness check of currentyly running unikernel failed \
+                       with: %s"
+                      err);
+                Middleware.http_response reqd
+                  ~title:"Error: Liveliness check failed"
+                  ~data:
+                    (`String
+                       ("Liveliness check of currentyly running unikernel \
+                         failed with error(s): " ^ err))
+                  `Bad_request))
```
hannesm, line 1558: and here's the very same error handling as above. this code duplication smells.

unikernel.ml
```diff
@@ -1411,87 +1422,140 @@ struct
...
+                        |> Yojson.Basic.to_string |> config_of_json)
+                    with
+                    | Ok cfg -> (
+                        process_change ~unikernel_name ~job
+                          ~to_be_updated_unikernel ~currently_running_unikernel
+                          cfg user store http_client `Update
+                        >>= function
+                        | Ok (unikernel_image, unikernel_cfg) -> (
+                            force_create_unikernel ~unikernel_name
+                              ~unikernel_image unikernel_cfg user albatross
```
hannesm, line 1453: here's the force-create of the unikernel where the arguments haven't changed -- and we don't perform liveliness checks now? why?

unikernel.ml
```diff
@@ -1411,87 +1422,140 @@ struct
...
+              ~dns_liveliness_address ~dns_liveliness_name http_client
+            >>= function
+            | Ok () -> (
+                process_change ~unikernel_name ~job ~to_be_updated_unikernel
+                  ~currently_running_unikernel cfg user store http_client
+                  `Update
+                >>= function
+                | Ok (unikernel_image, unikernel_cfg) -> (
+                    force_create_unikernel ~unikernel_name ~unikernel_image
+                      unikernel_cfg user albatross
```
hannesm, line 1507: here's the other force-create (we have modified the arguments), and now the liveliness checks are done.

utils.ml
```diff
@@ -179,6 +183,9 @@ let bytes_to_megabytes bytes =
 (*10 minutes for a rollback to be valid*)
 let rollback_seconds_limit = 600
 
+(* 60 seconds wait time to start a liveliness check*)
```
hannesm, line 186: is the comment up to date? "start a liveliness check"? now that there's an interval thereof?

liveliness_checks.ml
```diff
@@ -0,0 +1,121 @@
...
+            match check with
+            | `HTTP url -> "HTTP (" ^ url ^ "): " ^ err
+            | `DNS (address, domain_name) ->
+                "DNS (" ^ domain_name ^ ", " ^ address ^ "): " ^ err
+          in
+          perform_checks http_client (failed_desc :: failed) rest)
+
+let liveliness_checks ~http_liveliness_address ~dns_liveliness_address
+    ~dns_liveliness_name http_client =
+  Lwt.return
```
hannesm, line 55: why is this Lwt.return here? The whole `(...)` (until line 82) doesn't need to be inside the Lwt monad as far as I can see.

liveliness_checks.ml
```diff
@@ -0,0 +1,121 @@
...
+  | `HTTP address -> check_http http_client address
+  | `DNS (address, domain_name) ->
+      Error
+        (`Msg
+           (domain_name
+          ^ ": Can't handle DNS liveliness checks at the moment with address "
+          ^ address))
+      |> Lwt.return
+
+let rec perform_checks http_client failed = function
```
hannesm, line 31: If I understand this correctly, this is a fold over a list of checks? Would Lwt_list.fold_left_s work instead?

liveliness_checks.ml
```diff
@@ -0,0 +1,121 @@
+open Lwt.Infix
+
+let ( let* ) = Result.bind
+
+type checks = [ `HTTP of string | `DNS of string * string ]
+
+let check_http http_client base_url =
+  Utils.send_http_request http_client ~base_url >>= function
```
hannesm, line 8: You can use `>|=` instead of `>>=` and remove the `Lwt.return` in both error and ok case.

liveliness_checks.ml
```diff
@@ -0,0 +1,121 @@
...
+      Logs.err (fun m ->
+          m "http-liveliness-check: Error response of %s with error: %s"
+            base_url err);
+      Lwt.return
+        (Error
+           (`Msg
+              (base_url
+             ^ " :an error occured while performing a liveliness check on the \
+                http endpoint with error: " ^ err)))
+  | Ok _response -> Ok () |> Lwt.return
```
hannesm, line 19: not sure where I said this, but shouldn't the `_response` be more carefully checked (to be 200 - ok)?

liveliness_checks.ml
```diff
@@ -0,0 +1,121 @@
...
+
+let check_type http_client = function
```
hannesm, line 21: is there a timeout needed (for both HTTP and DNS)? As far as I'm concerned, we're waiting until we get an error or ok -- but what if there's something stuck? we should have a (rather small) timeout, otherwise mollymawk will be stuck waiting for a reply.

liveliness_checks.ml
```diff
@@ -0,0 +1,121 @@
...
+        Logs.info (fun m ->
+            m "Performing liveliness check of %s after %d second delay."
+              unikernel_name delay);
+        liveliness_checks ~http_liveliness_address ~dns_liveliness_address
+          ~dns_liveliness_name http_client
+        >>= function
+        | Error (`Msg err) -> aux_check (err :: failed_checks) rest
+        | Ok () -> aux_check failed_checks rest)
+  in
+  aux_check [] [ 1; 2; 5; 9; 14; 20; 27; 35; 44; 54 ]
```
hannesm, line 121: with this implementation, how long will be the maximum time? I guess since the delay will be slept, it is 211 seconds -- which is far more than the "liveliness_wait_period".

also, it looks like all errors are collected, so even if the third time the check(s) is/are successful, an error will be reported?

calling liveliness_checks in here means that the arguments are prepared several times -- and as far as I read the logic, even if the argument preparation fails, we get these error 10 times -- plus we wait for 211 seconds until we receive those errors!?

assets/main.js
hannesm: no need to act on this comment, but in the future it would be great if such things would be in a separate PR.

unikernel.ml
hannesm: same as above, please separate such refactorings into a separate PR in the future.

utils.ml
hannesm: also I think this is no longer used.

PizieDust: Yes you are right. With intervals, we don't need this anymore.

unikernel.ml
PizieDust: I was collecting several errors from the liveliness checks so I thought perhaps error(s) will be more appropriate. But we can change it

unikernel.ml
PizieDust: I've refactored this such that the liveliness_checks before the update is performed before config_or_none :)
thanks for catching that :)

unikernel.ml
PizieDust: I've pushed a commit which performs the liveliness_check before config_or_none and so it's only done once now

unikernel.ml
PizieDust: now there's a general liveliness_check before we do config_or_none :)

unikernel.ml
PizieDust: just refactored this, now we do the liveliness checks for both instances from one function

unikernel.ml
PizieDust: have refactored into a dedicated function `process_unikernel_update` which handles the update for when cfg is none and when it's some x and so the liveliness_checks will now be done for both without any duplication :)

liveliness_checks.ml
PizieDust: Thank you. this is now done using Lwt_list.fold_left_s :)

liveliness_checks.ml
PizieDust: Thank you. This works great

liveliness_checks.ml
PizieDust: This is handled in `Utils.send_http_request` and in the Status types of the package we have:

```
type successful = [
  | `OK
  | `Created
  | `Accepted
  | `Non_authoritative_information
  | `No_content
  | `Reset_content
  | `Partial_content
  ]
```
which is based on this rfc https://tools.ietf.org/html/rfc7231#section-6.3 and I think `Ok should be 200 but I just saw that we can match directly on the codes so I'll change it to match on the int code instead

unikernel.ml
PizieDust: Have refactored this to use just one dict entry for both

PizieDust: Have refactored this to use just one dict entry for both

liveliness_checks.ml
PizieDust: >with this implementation, how long will be the maximum time? I guess since the delay will be slept, it is 211 seconds -- which is far more than the "liveliness_wait_period".

Yes indeed, the maximum time with the current configuration will be 211 seconds. We no longer use a `liveliness_wait_period` since it's done in intervals. 

>also, it looks like all errors are collected, so even if the third time the check(s) is/are successful, an error will be reported?

I just refactored it such that it collects all errors, but now if there is a successful attempt, it returns on the successful attempt. I see the understanding being that with intervals, we are assuming that some services take long to start and so we keep checking in intervals and when we have a successful attempt, this means all the services are working and the liveliness_check is successful and we immediately return, and if it continuous to fail, we check till 211 seconds and then return all the failed attempts.

>calling liveliness_checks in here means that the arguments are prepared several times -- and as far as I read the logic, even if the argument preparation fails, we get these error 10 times -- plus we wait for 211 seconds until we receive those errors!?

I have separated the arguments preparation from the actual checking. This way if argument preparation fails, we immediately return, and if argument preparation succeeds, it's only done once and used for all the interval checks.

We are indeed doing a total of 10 attempts but with the current implementation, we can only reach 10 if the previous 9 have failed. If at any point, there's a successful check, we immediately return.

At the moment we have no implementation to display live notifications to the user. I will handle that in a separate PR so we can update the user for each failed attempt by toast notifications.

liveliness_checks.ml
PizieDust: We no longer require Lwt.return while preparing arguments as this is now handled in a dedicated function

liveliness_checks.ml
PizieDust: I agree. Most services uses 60 seconds to time out an HTTP request and 5 seconds to timeout a DNS check. 
We can use 15 seconds for out HTTP checks and maintain 5 seconds for DNS.

liveliness_checks.ml
```diff
@@ -0,0 +1,178 @@
+open Lwt.Infix
+
+let ( let* ) = Result.bind
+
+type checks = [ `HTTP of string | `DNS of string * string ]
+
+let http_timeout = 15 (* 15seconds timeout for http requests *)
```
hannesm, line 7:
```suggestion
let http_timeout = 15 (* 15 seconds timeout for http requests *)
```

liveliness_checks.ml
hannesm: Now that you mention it, I think any "successful" status should be taken as good enough for the liveliness check.

liveliness_checks.ml
hannesm: Hmm, now with the revised logic, I'm still struggling. We have timeouts now, that is great. But at the same time, we now wait in the worst case for 211 + 10 * 15 seconds = 361 seconds ~ 6 minutes... Now, this is what the web browser will need to wait if the unikernel does not start / work as expected. IMHO this is a tad too long.

What I propose:
- let's do less attempts or shorter intervals,
- let's keep the maximum time spend in "interval_liveliness_checks" to the 60 seconds we had initially in mind.

liveliness_checks.ml
```diff
@@ -0,0 +1,178 @@
...
+        in
+        Error (`Msg err_msg)
+
+  let perform_checks_with_timeout ~timeout stack http_client checks =
+    Lwt.pick
+      [
+        (perform_checks stack http_client checks >|= fun r -> `Result r);
+        (Mirage_sleep.ns (Duration.of_sec timeout) >|= fun () -> `Timeout);
+      ]
+
```
hannesm, line 102:
```suggestion
```

Let's remove this - I can't find any users thereof.

liveliness_checks.ml
```diff
@@ -0,0 +1,178 @@
+open Lwt.Infix
+
+let ( let* ) = Result.bind
+
+type checks = [ `HTTP of string | `DNS of string * string ]
+
+let http_timeout = 15 (* 15seconds timeout for http requests *)
+let dns_timeout = 5 (* 5 seconds timeout for dns requests *)
```
hannesm, line 8:
```suggestion
let timeout = 5 (* 5 seconds timeout for one check *)
```

liveliness_checks.ml
```diff
@@ -0,0 +1,178 @@
...
+            ( Mirage_sleep.ns (Duration.of_sec http_timeout) >|= fun () ->
+              `Timeout );
+          ]
+    | `DNS (address, domain_name) ->
+        Lwt.pick
+          [
+            (check_dns stack address domain_name >|= fun r -> `Result r);
+            ( Mirage_sleep.ns (Duration.of_sec dns_timeout) >|= fun () ->
+              `Timeout );
+          ]
```
hannesm, line 62:
```suggestion
  let check_type stack http_client typ =
    Lwt.pick
      [
        (Mirage_sleep.ns (Duration.of_sec timeout) >|= fun () -> `Timeout);
        (match typ with
        | `HTTP address -> check_http http_client address >|= fun r -> `Result r
        | `DNS (address, domain_name) ->
            check_dns stack address domain_name >|= fun r -> `Result r);
      ]
```

let's unify the timeout.g

liveliness_checks.ml
```diff
@@ -0,0 +1,178 @@
...
+          | delay :: rest -> (
+              Mirage_sleep.ns (Duration.of_sec delay) >>= fun () ->
+              Logs.info (fun m ->
+                  m "Performing liveliness check of %s after %d second delay."
+                    unikernel_name delay);
+              perform_checks stack http_client checks >>= function
+              | Error (`Msg err) -> aux_check (err :: failed_checks) rest
+              | Ok () -> Ok () |> Lwt.return)
+        in
+        aux_check [] [ 1; 2; 5; 9; 14; 20; 27; 35; 44; 54 ]
```
hannesm, line 176:
```suggestion
        aux_check [] [ 1; 1; 1; 2; 3; 3; 4; 5 ]
```

Let's use shorter intervals, with a maximum wait time of (1 + 1 + 1 + 2 + 3 + 3 + 4 + 5 = 20) + (8 * 5 = 40) = 60 seconds

liveliness_checks.ml
```diff
@@ -0,0 +1,178 @@
...
+    let validated_checks =
+      List.filter_map
+        (fun kind ->
+          match kind with
+          | `HTTP addr -> (
+              match addr with Some addr -> Some (`HTTP addr) | None -> None)
+          | `DNS (address, domain_name) -> (
+              match (address, domain_name) with
+              | Some address, Some domain_name ->
+                  Some (`DNS (address, domain_name))
```
hannesm, line 123: Here I'd do the "Domain_name.of_string domain_name" and "Domain_name.host" -- then we don't need them in "check_dns" (where the _exn variants are used).

At the same time, we can't really error out here, but we're in a "filter_map" and only return None or Some...

liveliness_checks.ml
hannesm: done now by the amended intervals and timeout

utils.ml
```diff
@@ -248,3 +252,25 @@ let switch_button ~switch_id ~switch_label html_content =
...
+  Http_mirage_client.request http_client ~follow_redirect:true
+    ~headers:[ ("Accept", "application/json") ]
+    url body_f body
+  >>= function
+  | Error (`Msg err) -> Lwt.return (Error (`Msg err))
+  | Error `Cycle -> Lwt.return (Error (`Msg "returned cycle"))
+  | Error `Not_found -> Lwt.return (Error (`Msg "returned not found"))
+  | Ok (resp, body) -> (
+      match resp.Http_mirage_client.status with
+      | `Code 200 -> Lwt.return (Ok body)
```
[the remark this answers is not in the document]
PizieDust: I imagine it to be the case, but to be on the safe side, we can accept all `OK responses

liveliness_checks.ml
PizieDust: I refactored a bit to handle this

utils.ml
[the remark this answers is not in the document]
PizieDust: I just tested and with builder_web we have the error: 

`
Error while fetching the current build info of hello-key-old with error: accessing https://builds.robur.coop/hash?sha256=fa34d22ededdd067c4d27ce655d1889bea56b209a346af83e349d7c168835897 resulted in an error: 200 OK
`

We can use the variants. 

which of these success variants below should we consider:

```
type successful = [
  | `OK
  | `Created
  | `Accepted
  | `Non_authoritative_information
  | `No_content
  | `Reset_content
  | `Partial_content
  ]
```

utils.ml
```diff
@@ -248,3 +252,25 @@ let switch_button ~switch_id ~switch_label html_content =
...
+    ~headers:[ ("Accept", "application/json") ]
+    url body_f body
+  >>= function
+  | Error (`Msg err) -> Lwt.return (Error (`Msg err))
+  | Error `Cycle -> Lwt.return (Error (`Msg "returned cycle"))
+  | Error `Not_found -> Lwt.return (Error (`Msg "returned not found"))
+  | Ok (resp, body) -> (
+      match resp.Http_mirage_client.status with
+      | `OK | `Created | `Accepted -> Lwt.return (Ok body)
+      | _ ->
```
hannesm, line 270:
```suggestion
      if Http_mirage_client.Status.is_successful resp.Http_mirage_client.status then
        Lwt.return (Ok body)
      else
```

hannesm:
Thanks a lot, this is ready to be merged (let's wait for CI).

Issue #73, opened 2024-10-18
Out of memory


reynir:
I triggered an out of memory exception while trying to deploy a ~17 MB unikernel with no arguments. The mollymawk unikernel itself runs with "only" 128 MB.

```
console 2024-10-18T09:54:08-00:00: 2024-10-18T09:54:08-00:00: [ERROR] [application] error exception Out of memory while processing request ((method "POST") (target "/unikernel/create") (version "HTTP/1.1") (headers 
console 2024-10-18T09:54:08-00:00:                                                  (("Priority" "u=0")
console 2024-10-18T09:54:08-00:00:                                                  ("Sec-Fetch-Site" "same-origin")
console 2024-10-18T09:54:08-00:00:                                                  ("Sec-Fetch-Mode" "cors")
console 2024-10-18T09:54:08-00:00:                                                  ("Sec-Fetch-Dest" "empty")
console 2024-10-18T09:54:08-00:00:                                                  ("Cookie" "molly_session=REDACTED")
console 2024-10-18T09:54:08-00:00:                                                  ("Connection" "keep-alive")
console 2024-10-18T09:54:08-00:00:                                                  ("Sec-GPC" "1")("DNT" "1")
console 2024-10-18T09:54:08-00:00:                                                  ("Origin" "https://mollymawk.robur.coop")
console 2024-10-18T09:54:08-00:00:                                                  ("Content-Length" "17617146")
console 2024-10-18T09:54:08-00:00:                                                  ("Content-Type" "multipart/form-data; boundary=---------------------------120320325341882077743984541787")
console 2024-10-18T09:54:08-00:00:                                                  ("Referer" "https://mollymawk.robur.coop/unikernel/deploy")
console 2024-10-18T09:54:08-00:00:                                                  ("Accept-Encoding" "gzip, deflate, br, zstd")
console 2024-10-18T09:54:08-00:00:                                                  ("Accept-Language" "en-US,en;q=0.5")
console 2024-10-18T09:54:08-00:00:                                                  ("Accept" "*/*")
console 2024-10-18T09:54:08-00:00:                                                  ("User-Agent" "Mozilla/5.0 (X11; Linux x86_64; rv:128.0) Gecko/20100101 Firefox/128.0")
console 2024-10-18T09:54:08-00:00:                                                  ("Host" "mollymawk.robur.coop"))))
```

reynir:
Trying again with a stripped unikernel image (9.1 MB) it OOM'd in a different place:
```
[...]

PR #123, opened 2025-04-11
Stream the unikernel image to albatross

hannesm:
Instead of fully reading and writing it to albatross, allow there to be a chunked stream. So, the unikernel image is written as data on the TLS flow in a chunked encoding (4 byte length followed by the chunk -- end of the data is marked by a 0 chunk).

Relies on https://github.com/robur-coop/albatross/pull/211, still needs the unikernel image to be streamed from the HTTP request to albatross / across the request handler.

//cc @reynir @PizieDust #73 https://github.com/robur-coop/albatross/pull/211 -- as mentioned, we need the http-multipart (and Utils.send_http_request) to deal with the chunking... esp. with the http-multipart thingy, I've no clue about the API.

### DIFF
--- a/albatross.ml
+++ b/albatross.ml
@@ -412,7 +412,7 @@ module Make (S : Tcpip.Stack.V4V6) = struct
       ]
     >|= fun output -> Result.map (fun r -> (r, output)) (decode_reply d)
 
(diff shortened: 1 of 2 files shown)


PizieDust:
Closing as moved to https://github.com/robur-coop/mollymawk/pull/137

PR #129, opened 2025-04-27
Better create form fixed

PizieDust:
This PR redesigns the form used to deploy unikernels. Previously all unikernel configurations were given as a a json string. Now the UI is more intuitive with a form element handling each part of the configuration: for instance selecting network interfaces and block devices, a toggle for fail_behaviour and a dropdown for cpuids etc.

### DIFF
--- a/account_usage.ml
+++ b/account_usage.ml
@@ -1,71 +1,11 @@
 let account_usage_layout policy unikernels blocks =
-  let deployed_unikernels = List.length unikernels in
-  let total_volume_used =
-    List.fold_left (fun total_size (_, size, _) -> total_size + size) 0 blocks
-  in
-  let total_free_space =
-    Option.value ~default:0 policy.Vmm_core.Policy.block - total_volume_used
-  in
-  let total_memory_used =
-    List.fold_left
-      (fun total_memory (_, unikernel) ->
-        total_memory + unikernel.Vmm_core.Unikernel.memory)
-      0 unikernels
-  in
-  let count_cpuid_usage =
-    let cpuid_count = Hashtbl.create (List.length unikernels) in
-    (* count usage of each cpuid in unikernels *)
-    List.iter
-      (fun (_, unikernel) ->
-        let cpuid = unikernel.Vmm_core.Unikernel.cpuid in
-        let count =
-          Hashtbl.find_opt cpuid_count cpuid |> Option.value ~default:0
-        in
-        Hashtbl.replace cpuid_count cpuid (count + 1))
-      unikernels;
-    (* Prepare list of all cpuids from policy *)
-    let policy_cpuids = Vmm_core.IS.elements policy.cpuids in
-    (* Generate the list with all policy.cpuids and their respective counts *)
-    List.map
-      (fun cpuid ->
-        let count =
-          Hashtbl.find_opt cpuid_count cpuid |> Option.value ~default:0
-        in
-        (cpuid, count))
-      policy_cpuids
-  in
-  let count_bridge_usage =
-    let bridge_count =
-      Hashtbl.create (Vmm_core.String_set.cardinal policy.bridges)
-    in
-    (* Initialize all policy bridges with a count of 0 *)
-    Vmm_core.String_set.iter
-      (fun bridge -> Hashtbl.replace bridge_count bridge 0)
-      policy.bridges;
-
-    (* Count each bridge usage from unikernel.bridges, but only if it's in policy.bridges *)
-    List.iter
-      (fun (_, unikernel) ->
-        List.iter
-          (fun { Vmm_core.Unikernel.host_device; _ } ->
-            (* Only count if the bridge is in policy.bridges *)
-            if Vmm_core.String_set.mem host_device policy.bridges then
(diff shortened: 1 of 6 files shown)


account_usage.ml
```diff
@@ -1,71 +1,11 @@
...
-    in
-    (* Sort the list by bridge name *)
-    List.sort (fun (b1, _) (b2, _) -> String.compare b1 b2) bridge_usage_list
+  let ( deployed_unikernels,
+        total_volume_used,
+        total_free_space,
+        total_memory_used,
+        count_cpuid_usage,
+        count_bridge_usage ) =
+    Utils.user_policy_usage policy unikernels blocks
```
hannesm, line 8: should this use a record instead of a six-tuple? I fear recording which element refers to what may be confusing since there are lots of `int` involved.

unikernel.ml
```diff
@@ -939,14 +939,24 @@ struct
...
     let now = Mirage_ptime.now () in
+    user_unikernels albatross user.name >>= fun unikernels ->
+    user_volumes albatross user.name >>= fun blocks ->
     generate_csrf_token store user now reqd >>= function
     | Ok csrf ->
+        let user_policy =
+          match Albatross.policy albatross ~domain:user.name with
+          | Ok p -> (
+              match p with Some p -> p | None -> Albatross.empty_policy)
+          | Error _ -> Albatross.empty_policy
```
hannesm, line 952: Hmm, why I understand that a default is good, the empty_policy leads to an error later -- since it won't allow any unikernel deployments.

So, I was wondering whether we should directly error out here instead!? And inform the user "you're not allowed to create any unikernels" / "couldn't retrieve resource policy from albatross".

unikernel_create.ml
```diff
@@ -1,4 +1,9 @@
-let unikernel_create_layout =
+let unikernel_create_layout ~(user_policy : Vmm_core.Policy.t) unikernels
+    (blocks : (Vmm_core.Name.t * int * bool) list) =
+  let _, _, _, total_memory_used, count_cpuid_usage, _ =
+    Utils.user_policy_usage user_policy unikernels blocks
+  in
```
hannesm, line 5: as mentioned above, I'd prefer a record.

hannesm: and now that only two values are used, I wonder whether it makes sense to split "Utils.user_policy_usage" into smaller functions?

unikernel_create.ml
```diff
@@ -1,4 +1,9 @@
-let unikernel_create_layout =
+let unikernel_create_layout ~(user_policy : Vmm_core.Policy.t) unikernels
```
hannesm, line 1: should we as well have a "[ ] force creation (destroys a unikernel already running with the same name)"?

unikernel_create.ml
```diff
@@ -16,37 +21,184 @@ let unikernel_create_layout =
...
+                        ]
+                      (List.map
+                         (fun (cpu_id, count) ->
+                           option
+                             ~a:[ a_value (string_of_int cpu_id) ]
+                             (txt
+                                ("CPU " ^ string_of_int cpu_id ^ " (used by "
+                               ^ string_of_int count ^ " unikernels)")))
+                         count_cpuid_usage);
+                  ];
```
hannesm, line 85: I'm not a HTML (form) expert, but should we by default have a selection that is the least used CPUid? Thus, by default nobody needs to click anything ;)

PizieDust:
> to me this looks great already, I've some minor comments. feel free to disregard them (or if it takes more effort, let's open separate issues for potential improvements)

thank you for the review. I think the latest commits address the various points you mentioned.

unikernel.ml
```diff
@@ -1613,30 +1613,50 @@ struct
...
-                  `Internal_server_error
-            | Ok (_hdr, res) -> (
-                match Albatross_json.res res with
-                | Ok res ->
-                    Middleware.http_response reqd ~title:"Success" ~data:res `OK
-                | Error (`String err) ->
-                    Middleware.http_response reqd ~title:"Error"
-                      ~data:(`String (String.escaped err))
-                      `Internal_server_error))
+            if bool_of_string force_create then (
```
hannesm, line 1627: we need to care about potential exceptions from bool_of_string -- this is user input.. maybe bool_of_string_opt?

hannesm:
From my memory from last week: how are arguments supplied now? Does space work as separator? Should we use a newline?

E.g. when I want to give the unikernel the arguments "--ipv4=1.2.3.4/24" and "--ipv4-gateway=1.2.3.1", what do I need to put into the field?

PizieDust:
> From my memory from last week: how are arguments supplied now? Does space work as separator? Should we use a newline?
> 
> E.g. when I want to give the unikernel the arguments "--ipv4=1.2.3.4/24" and "--ipv4-gateway=1.2.3.1", what do I need to put into the field?

It uses space as a separator now. So for the above case it will just be:
```
--ipv4=1.2.3.4/24 --ipv4-gateway=1.2.3.1
```

hannesm:
cool, let's merge this when the CI is fine.

PR #132, opened 2025-04-29
Prevent sliding cursors

PizieDust:
In the ui elements with a `-` and `+` buttons, manually typing a value reorders this as the cursor slides to the beginning. This PR fixes that.


hannesm:
thanks a lot.

PR #133, opened 2025-04-29
Treat json null the same as None, since that's how "config_to_json" works

hannesm:
//cc @PizieDust I tested this with our block device, and this works.


PizieDust (approved):

PizieDust:
Thank you @hannesm

PR #134, opened 2025-05-03
Better console output

PizieDust:
In this PR we introduce server sent events in polling the console output. 
Server sent events are just perfect for this sort of thing where data communication is unidirectional. 

https://github.com/user-attachments/assets/18c80170-e6ed-46fb-96ac-1e9c68bcef2c

### DIFF
--- a/albatross.ml
+++ b/albatross.ml
@@ -358,51 +358,57 @@ module Make (S : Tcpip.Stack.V4V6) = struct
     in
     split [] data 0 |> List.rev
 
-  let continue_reading name flow =
+  let rec continue_reading name f d flow =
     let open Lwt.Infix in
-    TLS.read flow >>= fun r ->
-    TLS.close flow >|= fun () ->
+    let bufs = split_many d in
+    let r =
+      List.fold_left
+        (fun acc d ->
+          match acc with
+          | Error () -> Error ()
+          | Ok () -> (
+              match decode d with
+              | Error s ->
+                  Logs.err (fun m ->
+                      m "albatross stop reading console %a: error %s"
+                        Vmm_core.Name.pp name s);
+                  Error ()
+              | Ok (_, `Data (`Console_data (ts, data))) -> f (ts, data)
+              | Ok (_, `Data (`Utc_console_data (ts, data))) -> f (ts, data)
+              | Ok w ->
+                  Logs.warn (fun m ->
+                      m
+                        "albatross unexpected reply, need console output, got \
+                         %a"
+                        (Vmm_commands.pp_wire ~verbose:false)
+                        w);
+                  Ok ()))
+        (Ok ()) bufs
+    in
     match r with
-    | Ok (`Data d) ->
-        let str = Cstruct.to_string d in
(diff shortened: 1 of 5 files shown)


assets/main.js
```diff
@@ -57,6 +57,27 @@ document.addEventListener('DOMContentLoaded', function () {
...
+		const unikernel_name = window.location.pathname.slice("/unikernel/info/".length);
+		const console_output = document.getElementById("console-output");
+		const events = new EventSource(`/api/unikernel/console/${unikernel_name}`);
+
+		events.onmessage = function (event) {
+			try {
+				const parsed = JSON.parse(event.data);
+				if (Array.isArray(parsed)) {
+					const newText = parsed.map(entry => `[${entry.timestamp}] ${entry.line}`).join("\n");
+					console_output.value = newText + "\n" + console_output.value;
```
hannesm, line 71: here I'd expect something to clip the output (to avoid infinite boxes).

assets/main.js
```diff
@@ -1175,3 +1196,4 @@ function areCharactersValid(s) {
 	}
 	return true;
 }
+
```
hannesm, line 1199: no need for this newline at the end of the file.

middleware.ml
```diff
@@ -118,6 +118,22 @@ let http_response ~title ?(header_list = []) ?(data = `String "") reqd
...
+    H1.Headers.(
+      add_list
+        (of_list [ ("Content-length", string_of_int (String.length data)) ])
+        header_list)
+  in
+  let response = H1.Response.create ~headers http_status in
+  let writer = H1.Reqd.respond_with_streaming reqd response in
+  let _ = H1.Body.Writer.write_string writer data in
+  let _ = H1.Body.Writer.flush writer (function () -> ()) in
+  H1.Body.Writer.close writer
```
hannesm, line 135: being a stream, I wouldn't close on the server side ever -- I'd be careful to receive client errors (i.e. the client closed the connection), and otherwise continue streaming...

albatross.ml
```diff
@@ -401,16 +401,8 @@ module Make (S : Tcpip.Stack.V4V6) = struct
 
   let console name tls_flow d =
     let open Lwt.Infix in
-    Lwt.choose
```
reynir, line 404: I think here you wanted Lwt.pick over Lwt.choose so the timeout is cancelled and doesn't close the flow.

hannesm:
I still am not 100% sure what the goal is here in terms of "better console output". If we have a unikernel that outputs console messages, and a browser that has the unikernel_info open -- what should the browser see?

As I understand this PR, the browser will see the last 20 console messages... and no updates when the unikernel emits new ones (in "continue_reading" (in contrast to the function name), the tls_flow is actually closed after a single read).

In order to get live updates - if we want that - my impression is:
- we need to open a HTTP stream with server-sent-events
- we need to open the albatross console_reading stream
- we need to keep the albatross stream alive as long as the browser is reading the HTTP stream

hannesm:
thanks a lot

PR #136, opened 2025-05-14
Make port number an argument

reynir:
I also changed the default port to 80 instead of 8080. I think that makes more sense for unikernels, but less so when testing with the unix target. Please feel free to change it to 8080.


PizieDust (approved):
Thank you @reynir

PR #137, opened 2025-05-19
Create unikernel with streaming.

PizieDust:
Streaming to albatross --Co-authored-by: hannes <hannes@mehnert.org>


albatross.ml
```diff
@@ -450,6 +450,37 @@ module Make (S : Tcpip.Stack.V4V6) = struct
...
+                  let rec send_data () =
+                    push () >>= function
+                    | None ->
+                        if !sent then (
+                          (* send trailing 0 byte chunk *)
+                          let buf = Cstruct.create 4 in
+                          Cstruct.BE.set_uint32 buf 0 0l;
+                          TLS.write tls_flow buf >>= function
+                          | Error _ -> assert false
+                          | Ok () -> Lwt.return_unit)
```
reynir, line 463: As far as I can tell we never set `sent := true`.

hannesm: well spotted (shouldn't there be a OCaml compiler warning)? This also explains why albatross doesn't deploy anything (it waits for the terminating zero-length chunk).

albatross.ml
```diff
@@ -450,6 +450,37 @@ module Make (S : Tcpip.Stack.V4V6) = struct
...
+                        else Lwt.return_unit
+                    | Some data -> (
+                        let buf = Cstruct.create (4 + String.length data) in
```
hannesm, line 466:
```suggestion
                        sent := true;
                        let buf = Cstruct.create (4 + String.length data) in
```

for fixing @reynir comment

reynir: I pushed a change that rewrites it to not use a reference

albatross.ml
reynir: It could in this case, but I that the analysis quickly gets hard if the `ref` escapes the scope.

albatross.ml
PizieDust: Thank you

albatross.ml
```diff
@@ -450,6 +450,38 @@ module Make (S : Tcpip.Stack.V4V6) = struct
...
               | Ok tls_flow -> (
+                  let write_one data =
+                    let buf = Cstruct.create (4 + String.length data) in
+                    Cstruct.BE.set_uint32 buf 0
+                      (Int32.of_int (String.length data));
+                    Cstruct.blit_from_string data 0 buf 4 (String.length data);
+                    TLS.write tls_flow buf >>= function
+                    | Error `Closed ->
+                        Logs.info (fun m -> m "tls got closed");
+                        assert false
```
hannesm, line 461: needs a proper error, no assert false

albatross.ml
```diff
@@ -450,6 +450,38 @@ module Make (S : Tcpip.Stack.V4V6) = struct
...
+                    | Error (`Write s) ->
+                        Logs.info (fun m ->
+                            m "tls write error which is: %a"
+                              S.TCP.pp_write_error s);
+                        assert false
```
hannesm, line 466: same, no assert false

albatross.ml
```diff
@@ -450,6 +450,38 @@ module Make (S : Tcpip.Stack.V4V6) = struct
...
+                    | Error _ -> assert false
```
hannesm, line 467: same

albatross.ml
```diff
@@ -450,6 +450,47 @@ module Make (S : Tcpip.Stack.V4V6) = struct
...
+                  in
+                  let send_data () =
+                    push () >>= function
+                    | None ->
+                      (* If we don't send data we don't send a trailing 0 byte chunk *)
+                      Lwt.return_unit
+                    | Some data -> (
+                        write_one data >>= fun () ->
+                        send_more_data ()
+                  in
```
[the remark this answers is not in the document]
reynir: I think it's less troublesome now that we don't have two meanings to `None`.

unikernel.ml
```diff
@@ -1253,9 +1298,9 @@ struct
   let force_create_unikernel ~unikernel_name ~unikernel_image
       (unikernel_cfg : Vmm_core.Unikernel.config) (user : User_model.user)
       albatross =
-    let unikernel_config = { unikernel_cfg with image = unikernel_image } in
-    Albatross.query albatross ~domain:user.name ~name:unikernel_name
-      (`Unikernel_cmd (`Unikernel_force_create unikernel_config))
+    let push () = Lwt.return (Some unikernel_image) in
```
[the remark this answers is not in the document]
MuluhGodson: For context, this function is mainly used during Unikernel updates ans the binary image is gotten from an endpoint on builds.robur.coop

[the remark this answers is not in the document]
PizieDust: I pushed a commit that creates a "fake stream" and feeds it into the push function.

unikernel.ml
```diff
@@ -1298,7 +1298,26 @@ struct
...
+          let len_to_read = min chunk_size (total_len - !current_pos) in
+          let chunk = String.sub unikernel_image !current_pos len_to_read in
+          current_pos := !current_pos + len_to_read;
+          push_chunks_into_stream (Some chunk);
+          Lwt.pause () >>= fun () -> loop ()
+      in
+      loop ()
+    in
+    Lwt.async feed_image_into_stream;
+    let push () = Lwt_stream.get image_data_stream in
```
reynir, line 1320: Ok, I see that this needs changes to the builder-web binary fetching code. I guess we can address that in another PR as it requires more changes.

unikernel.ml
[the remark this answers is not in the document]
hannesm: done in [229d2c6](https://github.com/robur-coop/mollymawk/pull/137/commits/229d2c6fb5999163ccf1a89f7c96695fd3e74876)

unikernel.ml
reynir: :clap: nice!

unikernel.ml
```diff
@@ -1384,12 +1366,39 @@ struct
...
-          ~path:
-            ("/job/" ^ job ^ "/build/" ^ to_be_updated_unikernel
-           ^ "/main-binary")
+        let data_stream, push_chunks = Lwt_stream.create () in
+        let push () = Lwt_stream.get data_stream in
+        let url =
+          Builder_web.base_url ^ "/job/" ^ job ^ "/build/"
+          ^ to_be_updated_unikernel ^ "/main-binary"
+        in
+        let f _response _acc chunk = Lwt.return (push_chunks (Some chunk)) in
```
reynir, line 1375: We should probably check the response status here already.

unikernel.ml
```diff
@@ -1384,12 +1366,39 @@ struct
...
+        Lwt.both
+          ( Http_mirage_client.request http_client ~follow_redirect:true
+              ~headers:[ ("Accept", "application/json") ]
```
reynir, line 1378: wait, why do we want to accept json here??

reynir: Likely, builder-web won't care but still

unikernel.ml
hannesm: indeed..

unikernel.ml
hannesm: see https://github.com/robur-coop/mollymawk/pull/143

PR #143, opened 2025-05-24
Address reviews from reynir in #137

hannesm:


reynir (approved):
:clap: thanks! looks good

PR #144, opened 2025-05-25
add a button to add policies to a user

### LINKED ISSUE #119: New user is created, there's no policy. Now, there's no link to create a policy.
Manually going to https://mollymawk.robur.coop/admin/u/policy/edit/<UUID> brings the UI to configure the policy.

~~There's another drawback in the UI, if entering numbers in the "allowed memory", these somehow turn in the wrong order (entering "2" "5" "6" results in "6" "5" "2").~~ (fixed by #132)

PizieDust:
fixes #119

### DIFF
--- a/user_single.ml
+++ b/user_single.ml
@@ -226,7 +226,7 @@ let user_single_layout (user : User_model.user) unikernels policy current_time =
                         a
                           ~a:
                             [
-                              a_href ("/admin/u/policy/edit/" ^ user.uuid ^ "");
+                              a_href ("/admin/u/policy/edit/" ^ user.uuid);
                               a_class
                                 [
                                   "border border-primary-500 \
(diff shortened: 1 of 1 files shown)


hannesm:
> While testing, I get this error from albatross: `policy policy: 5 unikernels 0,1,2 cpus 100 MB memory 150 MB block, bridges: is not smaller than root policy policy: 461168601 unikernels 0,1,2,3,4,5,6,7,8,9,10,11,12,13,14,15,16,17,18 cpus 16707 MB memory 302182 MB block: policy above allows bridge(s) , which is not a superset of` Will like to investigate that to see if it's not some bug in mollymawk

The newly defined policy has a set of bridges, namely the one which name is the empty string -- while the root policy does not have any bridge.

The printer for Policy in albatross emits a `bridges: ` only if the set of bridges is non-empty.

I don't know whether albatross should include the check to avoid bridges (and other devices) with the empty string. But mollymawk could as well be a bit more cautious.

Issue #157, opened 2025-10-02
HTTP return codes


hannesm:
From recent discussions:
- use 400 for malformed input (invalid unikernel name, invalid albatross name, invalid uuid)
- use 404 for resources that are not present (unikernel not found, albatross instance not found, ...)
- 500 for internal issues: failed to write to block device, failed to communicate with albatross (#153 has more nuances there)

right now we're mostly using 400, and sometimes 500 as well...

Issue #154, opened 2025-10-01
unikernel.ml error passing


hannesm:
I see some occurences of

```OCaml
            Middleware.http_response reqd ~title:"Error"
              ~data:
                (`String ("Couldn't convert name to albatross instance, " ^ err))
              `Bad_request)
```

but also:

```OCaml
                    Middleware.redirect_to_error
                      ~data:
                        (`String
                           ("An error occured while finding albatross instance "
                           ^ Configuration.name_to_str instance_name))
                      ~title:"Albatross Instance Error" ~api_meth:false
                      `Bad_request reqd ())
```

and another error handler:

```OCaml
        reply reqd ~content_type:"text/html"
          (Guest_layout.guest_layout ~page_title:"500 | Mollymawk"
             ~content:
               (Error_page.error_layout
                  {
                    code = 500;
                    success = false;
                    title = "Albatross Instance Error";
                    data =
                      `String
                        ("An error occured trying to find albatross instance "
                        ^ Configuration.name_to_str instance_name);
                  })
             ~icon:"/images/robur.png" ())
          `Bad_request
```

And I wonder whether we can uniform that somehow?

PR #158, opened 2025-10-02
Proper return status codes

PizieDust:
related to #157

### DIFF
--- a/albatross_instances.ml
+++ b/albatross_instances.ml
@@ -23,9 +23,7 @@ let select_instance (user : User_model.user) albatross_instances
              (fun (instance, (pol : Albatross.t)) ->
                let url =
                  a_href
-                   (Fmt.str "%s"
-                      (Middleware.construct_instance_redirect_url ~callback
-                         ~instance_name:(Configuration.name_to_str instance)))
+                   (Middleware.construct_instance_redirect_url callback instance)
                in
                match pol.error with
                | Some err ->
--- a/albatross_instances.ml
+++ b/albatross_instances.ml
@@ -23,8 +23,9 @@ let select_instance (user : User_model.user) albatross_instances
              (fun (instance, (pol : Albatross.t)) ->
                let url =
                  a_href
-                   (Fmt.str "%s?instance=%s" callback_url
-                      (Configuration.name_to_str instance))
+                   (Fmt.str "%s"
+                      (Middleware.construct_instance_redirect_url callback_url
+                         (Configuration.name_to_str instance)))
                in
                match pol.error with
                | Some err ->
--- a/albatross_instances.ml
+++ b/albatross_instances.ml
@@ -1,5 +1,5 @@
 let select_instance (user : User_model.user) albatross_instances
-    (callback_url : string) =
+    (callback : string) =
   Tyxml_html.(
     section
       ~a:[ a_id "instance-form"; a_class [ "p-4 bg-gray-50 my-1" ] ]
@@ -24,8 +24,8 @@ let select_instance (user : User_model.user) albatross_instances
                let url =
                  a_href
                    (Fmt.str "%s"
-                      (Middleware.construct_instance_redirect_url callback_url
-                         (Configuration.name_to_str instance)))
+                      (Middleware.construct_instance_redirect_url ~callback
+                         ~instance_name:(Configuration.name_to_str instance)))
                in
                match pol.error with
                | Some err ->
--- a/middleware.ml
+++ b/middleware.ml
@@ -97,10 +97,10 @@ let redirect_to_url ~url reqd ?(msg = "") () =
   H1.Reqd.respond_with_string reqd response msg;
   Lwt.return_unit
 
-let construct_instance_redirect_url ~callback ~instance_name =
-  match String.contains callback '?' with
-  | true -> callback ^ "&instance=" ^ instance_name
-  | false -> callback ^ "?instance=" ^ instance_name
+let construct_instance_redirect_url callback instance_name =
+  if String.contains callback '?' then
+    callback ^ "&instance=" ^ Configuration.name_to_str instance_name
+  else callback ^ "?instance=" ^ Configuration.name_to_str instance_name
 
 let redirect_to_instance_selector callback_link reqd ?(msg = "") () =
   redirect_to_url
--- a/middleware.ml
+++ b/middleware.ml
@@ -99,7 +99,8 @@ let redirect_to_url ~url reqd ?(msg = "") () =
 
 let construct_instance_redirect_url callback instance_name =
   let separator = if String.contains callback '?' then "&" else "?" in
-  Fmt.str "%s%sinstance=%" callback separator (Configuration.name_to_str instance_name)
+  Fmt.str "%s%sinstance=%" callback separator
+    (Configuration.name_to_str instance_name)
 
 let redirect_to_instance_selector callback_link reqd ?(msg = "") () =
   redirect_to_url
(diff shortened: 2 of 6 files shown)


unikernel.ml
```diff
@@ -1117,7 +1117,7 @@ struct
                          ~content:
                            (Error_page.error_layout (missing_policy_error ()))
                          ~icon:"/images/robur.png" ())
-                      `Internal_server_error)
+                      `Bad_request)
```
hannesm, line 1120: why is this a bad request now?

unikernel.ml
```diff
@@ -1126,7 +1126,7 @@ struct
                        (Error_page.error_layout
                           (missing_policy_error ~err:(Some err) ()))
                      ~icon:"/images/robur.png" ())
-                  `Internal_server_error)
+                  `Bad_request)
```
hannesm, line 1129: why is this a bad request? should Albatross_state.policy always return a "Ok (policy option)"? Or even an "Ok (Some policy)" for any user?

hannesm:
Thanks for your work on this. What I can tell is there are different HTTP status codes returned now in some situations. But from the diff I can't tell whether all cases are covered.

This maybe related to #154.

I'm also not entirely sure how it should look like, but maybe for a starter, how about replacing "get_instance_name" by a new function "albatross_instance", which:
- extracts from the query parameters the name
   - errors if the name is invalid
   - redirects to the selector if it does not exist
- directly picks the albatross instance
   -  if it doesn't exist, returns a not found error
-  then calls the function with the concrete albatross instance

Would that make the code more straightforward? This should make the functions called from the request_handler easier to follow. And if we then get some other functionality for POST to check the albatross-instance (and error out directly), we'd be able to condense (deduplicate) some more code. What do you think?

unikernel.ml
```diff
@@ -930,31 +930,31 @@ struct
...
               }
             in
             reply reqd ~content_type:"text/html"
-              (Guest_layout.guest_layout ~page_title:"401 | Mollymawk"
+              (Guest_layout.guest_layout ~page_title:"404 | Mollymawk"
                  ~content:(Error_page.error_layout error)
                  ~icon:"/images/robur.png" ())
-              `Unauthorized)
+              `Not_found)
     | Error (`Msg err) ->
```
reynir, line 944: Here the `error.title` is still "Unauthenticated" even though the error code was changes to not found / 404. There are other places with the same mistake.

reynir (commented):
I can recommend MDN's page about status codes: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status

unikernel.ml
PizieDust: So I looked at Albatross_state.policy and what can return an error is `Vmm_core.Name.path_of_string domain` and this looks like an error can happen if `domain` is the wrong format and hence why I changed it to a `Bad Request` rather than an `internal server error`. Let me know if you have a different opinion

unikernel.ml
```diff
@@ -2981,13 +2895,8 @@ struct
                   (extract_json_csrf_token (close_session store)))
         | path when String.starts_with ~prefix:"/volumes" path ->
```
hannesm, line 2896: I understand this hasn't changed in this PR, but why using the `String.starts_with` here? If I understand you correctly, you converted all the things to query parameters?

I'm asking since a few lines below you pass the string "/volumes" explicitly. So if there's something else in the path here (such as "/volumes/but/something/else", there'll be a mismatch / user confusion).

unikernel.ml
```diff
@@ -3067,14 +2970,11 @@ struct
...
                 in
-                match get_instance_name req with
-                | Ok instance_name ->
-                    authenticate ~check_admin:true store reqd
-                      (edit_policy store !albatross_instances uuid instance_name)
-                | Error err ->
-                    Middleware.redirect_to_error ~title:"Bad request"
-                      ~data:(`String ("Albatross name bad " ^ err))
-                      ~api_meth:false `Bad_request reqd ())
+
```
hannesm, line 2973: why is there an empty line?

unikernel.ml
```diff
@@ -3112,30 +3012,20 @@ struct
...
+                match get_unikernel_name req with
+                | Ok unikernel_name ->
                     authenticate store reqd
-                      (unikernel_info_one stack store !albatross_instances
-                         ~unikernel_name ~instance_name)
-                | Ok unikernel_name, _ ->
-                    Middleware.redirect_to_instance_selector
-                      ("/unikernel/info?unikernel=" ^ unikernel_name)
-                      reqd ()
+                      (albatross_instance "/unikernel/info"
```
hannesm, line 3018: see above, `path when String.starts_with` vs hardcoded "/uinkernel/info"

unikernel.ml
PizieDust: For `/volumes` it's always /volumes?instance=some_instance` so for this case, the callback will always just append `?instance` for us

albatross_instances.ml
```diff
@@ -23,8 +23,9 @@ let select_instance (user : User_model.user) albatross_instances
              (fun (instance, (pol : Albatross.t)) ->
                let url =
                  a_href
-                   (Fmt.str "%s?instance=%s" callback_url
-                      (Configuration.name_to_str instance))
+                   (Fmt.str "%s"
+                      (Middleware.construct_instance_redirect_url callback_url
+                         (Configuration.name_to_str instance)))
```
hannesm: `Fmt.str "%s"` is just not needed, no?

unikernel.ml
```diff
@@ -2784,13 +2786,18 @@ struct
         let req = H1.Reqd.request reqd in
         let path = Uri.(pct_decode (path (of_string req.H1.Request.target))) in
         let check_meth m f = if m = req.meth then f () else bad_request () in
+        let get_query_parameter name req =
+          match
+            Uri.get_query_param (Uri.of_string req.H1.Request.target) name
+          with
+          | Some param -> Ok param
+          | None -> Error "Couldn't find unikernel name in query"
```
hannesm, line 2794: The string includes `unikernel name`, but we should instead use `name` here!?

unikernel.ml
```diff
@@ -3012,10 +3009,11 @@ struct
                   (unikernel_info stack !albatross_instances))
         | path when String.starts_with ~prefix:"/unikernel/info" path ->
```
hannesm, line 3010: Would here `| "/unikernel/info" ->` work directly? If not, why not?

middleware.ml
```diff
@@ -97,6 +97,11 @@ let redirect_to_url ~url reqd ?(msg = "") () =
   H1.Reqd.respond_with_string reqd response msg;
   Lwt.return_unit
 
+let construct_instance_redirect_url callback instance_name =
+  if String.contains callback '?' then
+    callback ^ "&instance=" ^ Configuration.name_to_str instance_name
+  else callback ^ "?instance=" ^ Configuration.name_to_str instance_name
```
hannesm, line 103:
```suggestion
  let separator = if String.contains callback '?' then "&" else "?" in
  Fmt.str "%s%sinstance=%" callback separator (Configuration.name_to_str instance_name)
```

unikernel.ml
```diff
@@ -2948,11 +2860,16 @@ struct
...
+        | "/auth/verify" ->
             check_meth `GET (fun () ->
-                let token = String.sub path 19 (String.length path - 19) in
-                authenticate store reqd
-                  (email_verification (verify_email_token store token)))
+                match get_query_parameter "token" with
+                | Ok token ->
+                    authenticate store reqd
+                      (email_verification (verify_email_token store token))
+                | _ ->
```
hannesm, line 2869: a case matching everything else (`_ ->`) -> this is code smell and should be removed.

previously, the code was pretty lean, now it exploded in size. could there be a `get_query_parameter` function that errors directly (i.e. sends an error back) if the requested one is not found?

Anyways, maybe you could open an issue about it, and this can be addressed in a different PR?

PizieDust:
@hannesm thank you for the reviews

Issue #156, opened 2025-10-01
support albatross 2.6.0


hannesm:
that release has:
- `old_unikernel_info4`
- the type info and config now carry `add-name` field (which can be ignored, and always set to true)
- and a `startup` fields (int option), which should be visible in the UI and users should be able to set it (to a positive integer, by default they should use None)

hannesm:
fixed in #159

PR #159, opened 2025-10-05
support albatross 2.6

hannesm:
addresses #156

left TODO is a UI element for the `startup` field (an int option, by default None, the lower the earlier the unikernel is started (default is 50, it should be sufficient to go from 0 to 100)).

### DIFF
--- a/update_policy.ml
+++ b/update_policy.ml
@@ -45,7 +45,8 @@ let update_policy_layout (user : User_model.user) ~user_policy
               ~min_value:0
               ~max_value:
                 Vmm_core.Policy.(
-                  Option.value ~default:0 unallocated_resources.block + Option.value ~default:0 user_policy.block)
+                  Option.value ~default:0 unallocated_resources.block
+                  + Option.value ~default:0 user_policy.block)
               ~figure_unit:"MB" ~step:32 ~label':"Allowed Storage" ();
           ];
         hr ();


PizieDust:
the startup value, does it have a unit, like seconds?

assets/main.js
```diff
@@ -353,6 +353,7 @@ async function deployUnikernel(albatross_instance) {
 	const deployButton = document.getElementById("deploy-button");
 	const name = document.getElementById("unikernel-name").value;
 	const cpuid = document.getElementById("cpuid").value;
+	const startup = document.getElementById("startup-delay").innerText;
```
hannesm, line 356: It is no delay, please use "startup" or "startup_priority"

albatross_json.ml
```diff
@@ -285,12 +284,21 @@ let config_of_json str =
         (get "arguments" dict)
     with Failure s -> Error (`Msg ("expected strings as argv, error: " ^ s))
   in
+  let* startup =
+    Option.fold ~none:(Ok None)
+      ~some:(function
+        | `Int n when n >= 0 && n <= 100-> Ok (Some n)
+        | _ -> Error (`Msg "startup delay must be an integer between 1 and 100"))
+      (get "startup_delay" dict)
```
hannesm, line 292: no delay

unikernel_create.ml
```diff
@@ -85,86 +85,12 @@ let unikernel_create_layout ~(user_policy : Vmm_core.Policy.t) unikernels
...
-                          Unsafe.string_attrib "x-on:click" "count = count + 32";
-                        ]
-                      ~content:(i ~a:[ a_class [ "fa-solid fa-plus" ] ] [])
-                      ~btn_type:`Primary_outlined ();
-                  ];
+                Utils.increment_or_decrement_ui ~id:"unikernel-memory"
+                  ~max_value:memory_left ~min_value:32 ~figure_unit:"MB"
+                  ~label':"Memory";
+                Utils.increment_or_decrement_ui ~id:"startup-delay"
+                  ~max_value:100 ~min_value:1 ~figure_unit:"s"
```
hannesm, line 92: no unit.

unikernel_create.ml
```diff
@@ -85,86 +85,12 @@ let unikernel_create_layout ~(user_policy : Vmm_core.Policy.t) unikernels
...
+                  ~label':"Startup Delay";
```
hannesm, line 93: no delay

unikernel_create.ml
hannesm: also, please use 50 as default, as it is in albatross

albatross_json.ml
```diff
@@ -285,12 +284,21 @@ let config_of_json str =
         (get "arguments" dict)
     with Failure s -> Error (`Msg ("expected strings as argv, error: " ^ s))
   in
+  let* startup =
+    Option.fold ~none:(Ok None)
+      ~some:(function
+        | `Int n when n >= 0 && n <= 100-> Ok (Some n)
```
hannesm, line 290: if n = 50 then None else Some n

unikernel_index.ml
```diff
@@ -32,9 +41,7 @@ let instance_unikernels instance_name albatross_instance_unikernels current_time
                     ];
                 ]
               [
-                txt
-                  (match unikernel.Vmm_core.Unikernel.typ with
-                  | `Solo5 -> "solo5");
+                txt (string_of_int (Option.value ~default:50 unikernel.startup));
```
hannesm, line 44: is it worth to annotate the default (none) case as "50 (default)"?

utils.ml
```diff
@@ -155,16 +155,19 @@ let button_component ~attribs ~content ~btn_type ?(extra_css = "") () =
 
 (** A UI with two buttons, one for decrementing a value and the other for
     incrementing the value*)
-let increment_or_decrement_ui ~max_value ~min_value ~figure_unit ~id ~label' =
+let increment_or_decrement_ui ~max_value ~min_value ?(default_value = 0)
+    ?(figure_unit = "") ~id ~label' () =
   Tyxml_html.(
```
hannesm, line 160: so you never check that max is bigger than min?

what if min is -10, max 10, and default should indeed be 0? I suggest to have no default value for default to distinguish the valid use of 0 from nothing supplied

unikernel_index.ml
```diff
@@ -4,6 +4,9 @@ let instance_unikernels instance_name albatross_instance_unikernels current_time
     List.map
       (fun (name, unikernel) ->
         let name = Option.value ~default:"no name" (Vmm_core.Name.name name) in
+        let startup =
+          Option.value ~default:50 unikernel.Vmm_core.Unikernel.startup
```
hannesm, line 8: sorry for being picky, but why not:
Option.value ~default:"50 (default)"
(Option.map string_of_int unikernel.startup)

PizieDust: Thanks. I tried at frist but didn't know to combine `Option.map` with `string_of_int` inside the `Option.value`. I think this is much better

hannesm:
thanks

Issue #153, opened 2025-09-26
New Error Semantics for Mollymawk/Albatross


PizieDust:
# Error Handling and failure semantics for Albatross instances

## 1. Summary
_tldr_
The current implementation of error handling for albatross instances is limited to an `error: string option`. This is a good first step to prevent failing instances from silently disappearing from the UI, but it has several shortcomings:
- **opaqueness**: a single string doesn't provide enough context to determine the nature of the error.
- **lack of history**: it only stores the most recent error, losing valuable context if a sequence of different failures occurs.
- **lack of strategy**: we can't distinguish between a temporary network blip and a permanent misconfiguration. This prevents us from implementing intelligent recovery mechanisms.

here we (@hannesm  and I)  propose a more structured way to track, categorize, and recover from failures in albatross instances. The goal is to provide a clear history of failures for adminis and enable automated recovery where possible

## 2. Current Implementation

The current state for an Albatross instance includes a simple optional string for errors:

```ocaml
type t = {
  mutable policies : Vmm_core.Policy.t Vmm_trie.t;
  configuration : Configuration.t;
  error : string option;
}
```

## 3. Proposed Error Categorization

We propose categorizing errors into distinct classes. This table defines the category for each *new* error event that is added to the log.

| Category | Description | Examples | Responsible Party |
| :--- | :--- | :--- | :--- |
| **1. User-Provoked** | Errors caused by invalid user input or requests correctly rejected by Albatross. | Malformed unikernel image or wrong unikernel name format. | **User** |
| **2. Persistent Config** | Unrecoverable errors that occur during initialization. | Invalid TLS cert, incorrect IP address. | **Admin** |
| **3. Transient Network** | Errors indicating a temporary inability to communicate. | TCP timeout, Albatross daemon restarting. | **Mollymawk (Automated)** |
| **4. Incompatibility** | Errors due to a protocol or version mismatch between Mollymawk and Albatross. | Albatross too old or too new. | **Admin** |
| **5. Internal Mollymawk** | An unexpected error or exception within Mollymawk itself. | Storage device unresponsive | **Admin** |

## 4. Data types

To support an error history, a new `status` type will be included to hold a **list** of `error_info` records, replacing the `error: string option`

```ocaml
(* To be defined in a new submodule, e.g., Albatross.Instance_status *)

(** Information captured for a single failure event. *)
type error_info = {
  timestamp : Ptime.t;      
  category : [                
    | `Configuration
    | `Transient
[...]

PR #164, opened 2025-10-20
Detailed status logs for albatross instances

PizieDust:
this PR is related to https://github.com/robur-coop/mollymawk/issues/153

Here we define a new module for the status of each albatross instance.
- Admins can see full error logs and retry connecting to an albatross instance.
- Normal users can also retry connecting without seeing full error logs.
- Once an albatross instance goes online, it's error logs are cleared.

### DIFF
--- a/albatross.ml
+++ b/albatross.ml
@@ -577,7 +577,14 @@ module Make (S : Tcpip.Stack.V4V6) = struct
                 | Ok () -> (
                     TLS.read tls_flow >>= fun r ->
                     match r with
-                    | Ok (`Data d) -> f tls_flow (Cstruct.to_string d)
+                    | Ok (`Data d) -> (
+                        f tls_flow (Cstruct.to_string d) >|= function
+                        | Ok _ as o -> o
+                        | Error msg as e ->
+                            t.status <-
+                              Status.update t.status
+                                (Status.make `Incompatible msg);
+                            e)
                     | Ok `Eof ->
                         TLS.close tls_flow >|= fun () ->
                         let err =


albatross.ml
```diff
@@ -1,9 +1,42 @@
...
+      | `Internal -> "internal"
+    in
+    Fmt.pf ppf "[%a] %s: %s" Ptime.pp timestamp category_str details
+end
+
 type t = {
   mutable policies : Vmm_core.Policy.t Vmm_trie.t;
   configuration : Configuration.t;
-  error : string option;
+  status : Instance_status.t;
```
hannesm: likely needs to be mutable.

PizieDust:
> What needs to be covered is who's able to see errors, 

Only users with admin accounts

> and who's able to "try recovery"? As example, if there's a network failure reaching one albatross during deployment of user A, would user B see this error? Would A or B be able to try again deploying to that albatross instance?

If user A tried deploying and had an error, this will display on their UI (like a toast notification) and they will be on the page where they tried deploying from. Here they can just click the `Deploy` button again. 

User B will not be able to see this error, unless they are an admin account and can see the entire error log of this albatross instance. User B will not be able to try deploying what User A tried to deploy as this is only available on the UI of User A.

unikernel.ml
```diff
@@ -379,40 +399,27 @@ struct
     Albatross_state.query stack state ~domain:user_name (`Block_cmd `Block_info)
     >|= function
     | Error msg ->
-        Logs.err (fun m ->
-            m "Error querying volumes on '%s': %s"
-              (Configuration.name_to_str state.configuration.name)
-              msg);
+        log_albatross_status state `Configuration (`String msg);
```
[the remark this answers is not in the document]
PizieDust: and which of these should be classify as Degraded and which should we classify as Offline.
for me, we can classify all transient errors as Degraded and anything Configuration/Incompatible should be classified as Offline?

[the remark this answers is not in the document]
PizieDust: > Now, I've a hard time to understand what "Tls.Config.client error" means -- I mean when can this happen, and is it transient or permanent? 

Tls.Config.client I think fails if the provided certificates (from Configuration.t) are invalid. So I think we can classify it as a `Configuration` error

albatross.ml
```diff
@@ -1,9 +1,41 @@
...
+    in
+    Fmt.pf ppf "[%a] %s: %s" Ptime.pp timestamp category_str details
+end
+
 type t = {
   mutable policies : Vmm_core.Policy.t Vmm_trie.t;
   configuration : Configuration.t;
-  error : string option;
+  mutable status : Status.t;
 }
```
hannesm: maybe here a `let set_online t = t.status <- Status.Online` is suitable?

albatross.ml
```diff
@@ -585,17 +620,20 @@ module Make (S : Tcpip.Stack.V4V6) = struct
...
-                  (Vmm_commands.pp_wire ~verbose:false)
-                  w);
+            let err =
+              Fmt.str "Unexpected policy reply for %s: %a"
+                (Configuration.name_to_str configuration.name)
+                (Vmm_commands.pp_wire ~verbose:false)
+                w
+            in
+            Logs.warn (fun m -> m "%s" err);
+            let error = Status.make `Internal err in
```
hannesm: It is a bit unclear to me what `` `Internal `` actually means -- since the issue here is that the remote albatross replied with something unexpected --> this means either Mollymawk needs to be updated or downgraded (or the remote albatross needs to be downgraded or updated) --> since the remote administrator may influence the behaviour, I lean towards making this transient, but as said, I'm not sure.

albatross.ml
```diff
@@ -606,11 +644,15 @@ module Make (S : Tcpip.Stack.V4V6) = struct
...
+                (Fmt.str "Initialisation error: raw_qeury failed for %s with %s"
+                   (Configuration.name_to_str configuration.name)
+                   str)
+            in
             let state =
               {
-                error =
-                  Some
-                    (Fmt.str "Failed to query this instance with error: %s" str);
+                status = Offline [ error ];
```
hannesm: Do we have recovery from `Offline` (is there a button to try communicating with this albatross)?

unikernel.ml
```diff
@@ -379,40 +410,31 @@ struct
...
+        blocks
     | Ok reply ->
-        Logs.err (fun m ->
-            m "Expected a block info reply from '%s', received %a"
-              (Configuration.name_to_str state.configuration.name)
-              (Vmm_commands.pp_wire ~verbose:false)
-              reply);
+        update_albatross_status state
+          (Some (`Transient, `Res (reply, "block info")));
         []
```
hannesm: Hmm, can't we do the albatross status updates directly in `Albatross_state.query`? There we also know all the fine details...

albatross.ml
```diff
@@ -530,28 +585,41 @@ module Make (S : Tcpip.Stack.V4V6) = struct
...
+                        t.status <-
+                          Status.update t.status (Status.make `Transient err);
+                        Error err)
+                | Error err ->
+                    t.status <-
+                      Status.update t.status (Status.make `Transient err);
+                    Lwt.return_error err)))
 
   let init stack (configuration : Configuration.t) =
     let open Lwt.Infix in
```
hannesm: if you create a `state` at the beginning of this function...

albatross.ml
```diff
@@ -565,37 +633,37 @@ module Make (S : Tcpip.Stack.V4V6) = struct
...
-                  w);
+            let err =
+              Fmt.str "Unexpected policy reply for %s: %a"
+                (Configuration.name_to_str configuration.name)
+                (Vmm_commands.pp_wire ~verbose:false)
+                w
+            in
+            Logs.warn (fun m -> m "%s" err);
+            let error = Status.make `Internal err in
             let state =
```
hannesm: here you'd `state.status <- Status.update Online error;` (instead of the `let state = ...`)

albatross.ml
```diff
@@ -606,11 +674,15 @@ module Make (S : Tcpip.Stack.V4V6) = struct
                 m "init: raw_query failed for %s: %s"
                   (Configuration.name_to_str configuration.name)
                   str);
+            let error =
+              Status.make `Configuration
+                (Fmt.str "Initialisation error: raw_qeury failed for %s with %s"
+                   (Configuration.name_to_str configuration.name)
+                   str)
+            in
             let state =
```
hannesm: see above, `state.status <- Offline [error];`, no `let state = ... ` anymore.

PizieDust:
> It looks nice, but why do we now have "settings" and "albatross_instances" -- what is the purpose of two pages listing the albatross instances? (I'm fine if there's a deeper meaning / concrete issue to solve with "albatross instances" that "settings" should not be aware of -- but I'm a bit lost why there are two pages that are accessible by administrator only that list all albatross instances and allow you to do something with them.)

So for the `settings` page: 
- Here we handle albatross configurations 
- Here we have the button which takes the admin to the status page of the instance (where they can see errors)
- Here is also where will add functionality to configure email

For the `albatross_instances` page:
- This is the main redirect page, which is displayed before the user can perform an action with albatross. For instance, to deploy a new unikernel, the user is first presented the albatross_instances page and here they can select the instance they are interested in, same with creating or downloading block devices etc.
- This is where the user also has an opportunity to retry connecting to an instance. 

In a future PR, we will refactor this to get rid of the albatross_instances page, and just have a dropdown select where the user can dynamically select the instance the want and the page updates dynamically, and they don't need to have an intermediary page: see https://github.com/robur-coop/mollymawk/pull/167 (for a first implementation)

unikernel.ml
```diff
@@ -963,21 +949,50 @@ struct
...
+        albatross.status <-
+          Albatross.Status.update albatross.status
+            (Albatross.Status.make `Transient "Re-initialization failed.");
+        Middleware.http_response reqd
+          ~data:(`String "Re-initialization failed. See error logs")
+          `Internal_server_error
+    | Offline _ ->
+        albatross.status <-
+          Albatross.Status.update albatross.status
+            (Albatross.Status.make `Configuration "Re-initialization failed.");
```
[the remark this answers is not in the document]
PizieDust: Thanks for catching that. Status updates are updated in the `Albatross.init` function which also calls `raw_query` and hence we don't need to update here anymore.

PR #165, opened 2025-10-27
Uniform names

### LINKED ISSUE #160: Appropriate types for user name and unikernel name
We pass both to albatross, so they're restricted to what albatross accepts -- namely a string where Vmm_core.Name.valid_label passes.

We should enforce this in albatross as well by using proper types, and thus catching invalid names early (to avoid they're leading to errors deep inside the business logic which is already complex).

See https://github.com/robur-coop/mollymawk/pull/158#discussion_r2404619004 for further reference.

PizieDust:
fixes #160 

The type for user names is changed from `string` to `Vmm_core.Name.Label.t`, making it use the same validation rules Albatross expects

### DIFF
--- a/albatross.ml
+++ b/albatross.ml
@@ -56,12 +56,11 @@ module Make (S : Tcpip.Stack.V4V6) = struct
       }
 
   let policy ?domain t =
-    let ( let* ) = Result.bind in
-    let* path =
-      match domain with
-      | None -> Ok Vmm_core.Name.root
-      | Some domain ->
-          Ok (Vmm_core.Name.make_of_path (Vmm_core.Name.Path.of_label domain))
+    let path =
+      Option.value ~default:Vmm_core.Name.root
+        (Option.map
+           (fun d -> Vmm_core.Name.make_of_path (Vmm_core.Name.Path.of_label d))
+           domain)
     in
     Ok (Vmm_trie.find path t.policies)
 


albatross.ml
```diff
@@ -667,19 +655,16 @@ module Make (S : Tcpip.Stack.V4V6) = struct
...
-        (* now we tell albatross about it, using a command for throwing it away *)
-        (* note that the 'certs' / 'gen_cert' uses the policies for intermediate certificates *)
-        query stack t ~domain (`Unikernel_cmd `Unikernel_info) >|= function
-        | Ok _ -> Ok (name, policy)
-        | Error msg ->
-            Logs.warn (fun m -> m "error updating policies: %s" msg);
-            t.policies <- old_policies;
-            Error msg)
+    let old_policies = t.policies in
+
```
hannesm: again, a spurious newline appears...

hannesm:
now that albatross 2.6.1 is released (and #169 addressed the API changes), this needs to be rebased. thanks.

albatross.ml
```diff
@@ -60,12 +60,8 @@ module Make (S : Tcpip.Stack.V4V6) = struct
...
       | None -> Ok Vmm_core.Name.root
-      | Some domain -> (
-          match Vmm_core.Name.Path.of_string domain with
-          | Error (`Msg msg) ->
-              Error
-                (Fmt.str "albatross: domain %s is not a path: %s" domain msg)
-          | Ok path -> Ok (Vmm_core.Name.make_of_path path))
+      | Some domain ->
+          Ok (Vmm_core.Name.make_of_path (Vmm_core.Name.Path.of_label domain))
     in
```
hannesm: This is basically now something that will always succeed, no?

```OCaml
let path = Option.value ~default:Vmm_core.Name.root (Option.map (fun d -> Vmm_core.Name.make_of_path (Vmm_core.Name.Path.of_label d)) domain) in
```

Or is there something I'm missing?

PR #166, opened 2025-10-29
Only use String.escaped for user supplied data

### LINKED ISSUE #163: Input handling, and the use of String.escaped
We should uniformly handle input, and printing (parts of) input data to logs or sending it back to the HTTP client. There are some occasions where String.escaped is used, but it would be great to have this more uniform.

PizieDust:
closes #163 

In this PR we remove a bunch of `String.escaped` where it's redundant and only instances where we have to return user input do we uses.

### DIFF
--- a/utils.ml
+++ b/utils.ml
@@ -106,7 +106,7 @@ module Status = struct
         ("success", `Bool status.success);
         ("data", status.data);
       ]
-    |> Json.to_string
+    |> Yojson.Basic.to_string
 end
 
 let csrf_form_input csrf =


unikernel.ml
```diff
@@ -2105,14 +2082,10 @@ struct
...
-                                (`String
-                                   ("error with root policy: "
-                                  ^ String.escaped err))
+                              ~data:(`String ("error with root policy: " ^ err))
                               `Internal_server_error)
                     | Error (`Msg err) ->
-                        Middleware.http_response reqd
-                          ~data:(`String (String.escaped err))
+                        Middleware.http_response reqd ~data:(`String err)
                           `Bad_request)
```
hannesm: here I'm a bit unsure -- so Albatross_json.policy_of_json in the error case uses `Yojson.Basic.to_string` -- and I'm not sure whether this is safe (in terms of injection attacks).

unikernel.ml
```diff
@@ -1805,8 +1783,7 @@ struct
                     match Albatross_json.res res with
                     | Ok res -> Middleware.http_response reqd ~data:res `OK
                     | Error (`String err) ->
-                        Middleware.http_response reqd
-                          ~data:(`String (String.escaped err))
+                        Middleware.http_response reqd ~data:(`String err)
```
hannesm: see above

unikernel.ml
```diff
@@ -233,9 +233,7 @@ struct
   let extract_json_csrf_token f token_or_cookie user reqd =
     extract_json_body reqd >>= function
     | Error (`Msg err) ->
-        Middleware.http_response reqd
-          ~data:(`String (String.escaped err))
-          `Bad_request
+        Middleware.http_response reqd ~data:(`String err) `Bad_request
```
hannesm: with the state of unikernel.ml after this commit, there are three String.escaped left -- and two of them are in exactly this case -- (here, extract_json_body calls `Yojson.Basic.of_string`, and the resulting `Json_error msg` -- should `msg` be escaped or not)? Since I don't understand yojson too well, I'm not sure...

but we should treat this uniform :)

hannesm:
this looks excellent, we're close :) thanks for the cleanups!

hannesm:
I've two questions:
- would you mind to rebase (there's a merge conflict)?
- you reported some time ago that this breaks configuration data stored on disk -- is this better now?

PizieDust:
> I've two questions:
> 
> * you reported some time ago that this breaks configuration data stored on disk -- is this better now?

The problem was that when we escape a json string before dumping it on the disk, to unserialize the json will produce a string with more quotes and this is problematic, for instance matching passwords, or checking if a cookie is valid. 
So i remove the escaping from the data we write to the block device, and mostly focused on the data we send back to users and what we get from them.

hannesm:
Thanks for your work, I still have the question whether the exception raised by Yojson.Basic.of_string is fine to transport via HTTP (or needs to be wrapped into a String.escaped)?

You introduce Utils.Json.from_string -- but this is just an alias to Yojson.Basic.from_string -- I wonder whether you'd want to put the exception stuff in there as well:
```OCaml
let from_string str =
  try Ok (Yojson.Basic.from_string data)
  with Yojson.Json_error s -> Error (`Msg (String.escaped s)) (* and here I'm not sure whether we need a String.escaped or not *)
```

This would certainly have an impact on the call-site (where now no `try .. with ..` would be needed).

PizieDust:
```ocaml
| "/check" ->
            check_meth `GET (fun () ->
                reply reqd ~content_type:"text/html"
                  "<script>console.log('a test')</script>" `OK)
```
- `a test` is printed in the console
---

```ocaml
| "/check" ->
            check_meth `GET (fun () ->
                Middleware.http_response ~api_meth:false reqd
                  ~data:
                    (Yojson.Basic.from_string
                       "<script>console.log('a test')</script>")
                  `OK)
```
- Mollymawk fails with `Segmentation fault (core dumped)`

---

```ocaml
| "/check" ->
            check_meth `GET (fun () ->
                Middleware.http_response ~api_meth:false reqd
                  ~data:
                    (`String
                       "<script>console.log('a test')</script>")
                  `OK)
```
- `a test` is printed in the console


I think we should escape yojson stuff

PR #167, opened 2025-10-30
Add a dropdown select for albatross instances

PizieDust:
This PR adds a dropdown to filter unikernels by albatross instance.

image

### DIFF
--- a/unikernel_index.ml
+++ b/unikernel_index.ml
@@ -138,7 +138,7 @@ let unikernel_index_layout unikernels_by_albatross_instance current_time =
                   ~a:[ a_class [ "font-bold text-gray-700" ] ]
                   [
                     txt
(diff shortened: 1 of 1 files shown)


unikernel_index.ml
```diff
@@ -3,10 +3,7 @@ let instance_unikernels instance_name albatross_instance_unikernels current_time
   Tyxml_html.(
     List.map
       (fun (name, unikernel) ->
-        let name =
-          Option.value ~default:"no name"
-            (Option.map Vmm_core.Name.Label.to_string (Vmm_core.Name.name name))
-        in
+        let name = Vmm_core.Name.to_string name in
```
hannesm: I don't understand this change. Doesn't it then again display ":robur-next" instead of "robur-next"?

PR #170, opened 2025-12-01
Add email functionality to Mollymawk

PizieDust:
This PR adds email functionality into mollymawk using the sendmail package. 
Here we need an IP, Port and Sender email. 

On our Settings page we now have tabs for Albatross configuration and a second tab for Email configuration.

### DIFF
--- a/unikernel.ml
+++ b/unikernel.ml
@@ -34,9 +34,7 @@ struct
     | Ok path -> [ Colombe.Forward_path.Forward_path path ]
     | Error (`Msg e) ->
         Logs.err (fun m ->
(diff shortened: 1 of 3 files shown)


dinosaure:
Seems good, I'm not sure you should always use `Mrmime.Mailbox.to_string` and it's probably better to use `emile` here. `mrmime` tries to respect the 80-columns rule and can insert some spurious newlines and spaces for that. `emile` does not do that for instance. So if you want to serialize an email address for a HTML page, for instance, it's probably better to use `Emile.to_string`.

PizieDust:
Thanks for the reviews @dinosaure

Issue #100, opened 2024-12-11
fine-grained permissions for API tokens


hannesm:
this may be useful at some point

PR #188, opened 2025-12-09
check auth token for some endpoints

PizieDust:
While writing documentation for the rest API and testing them, I noticed the API methods for tokens management was missing the `check_token` flag. This PR corrects that.


hannesm:
I'm by no means an expert, but this means that once I have a token I can update it, and create new tokens? Given that we don't have any fine-grained permissions #100, this makes a token as valuable as having the username & password.

PR #194, opened 2025-12-15
Update to cachet.0.0.3

reynir:
This requires albatross.2.6.2 https://github.com/ocaml/opam-repository/pull/29097 as albatross.2.6.1 is not compatible with cachet.0.0.3. I did not update the version constraint as it's internal to albatross the dependency on cachet.

### DIFF
--- a/albatross.ml
+++ b/albatross.ml
@@ -129,8 +129,7 @@ module Make (S : Tcpip.Stack.V4V6) = struct
   let manifest_devices_match ~bridges ~block_devices binary =
     let cachet =
       let map () ~pos len =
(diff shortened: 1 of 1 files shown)


reynir:
Right, yes, let's do that then.

In https://git.robur.coop/robur/ocaml-solo5-elftool/pulls/3 I had a streaming implementation that worked about half of the cases I tested against. The issue is that some tables are not in the order we want to look up things in. I managed to find a heuristic that works in some of the cases, but I doubt there is something we can do that works all the time. More context in https://github.com/robur-coop/ocaml-solo5-elftool/issues/3.

PR #203, opened 2026-01-23
Module for autoscaling logic of unikernels

PizieDust:
This is the first step to automatic scaling of unikernels. This pr adds a new module `Autoscaler` which is meant to handle all the logic involved with checking the load of a unikernel and determining it's state. We have 4 states:

- `Overloaded:` we should spawn a new clone.
- `Pending:` usage is high but we are still monitoring to be sure it wasn't a single incidental spike.
- `Cooldown:` we spawned a new unikernel and now we'll wait for sometime and monitor usage again.
- `Normal:`  all is okay with no usage spikes

We also have some static values which determine the behaviour of the autoscaler:
- cpu is checked every 5 minutes
- if CPU ≥ 90% for 5 consecutive checks i.e overload detected (~25 minutes sustained load)
- after spawning a new clone:
     - ignore overload signals for 10 minutes

### DIFF
--- a/autoscaler.ml
+++ b/autoscaler.ml
@@ -1,4 +1,4 @@
-let poll_interval = 300.0 (* check every 5 minutes *)
+let poll_interval = Duration.of_min 5 (* check every 5 minutes *)
 let trigger_ticks = 5
 (* number of times a unikernel is checked and is overloaded before spawning a new clone i.e trigger_ticks x poll_interval determines how load a unikernel is overloaded for before we spawn. Currently this will be 25 minutes. *)
 
--- a/autoscaler.ml
+++ b/autoscaler.ml
@@ -19,7 +19,7 @@ module Cpu_monitor = struct
     let sys_t = rusage_to_float r.stime in
     user_t +. sys_t
 
-  type t = { mutable last_cpu_time : float; mutable last_wall_time : Ptime.t }
+  type t = { last_cpu_time : float; last_wall_time : Ptime.t }
 
   let create now (initial_rusage : Vmm_core.Stats.rusage) =
     { last_cpu_time = get_total_cpu_time initial_rusage; last_wall_time = now }
@@ -30,8 +30,8 @@ module Cpu_monitor = struct
     let cpu_delta = curr_cpu_time -. t.last_cpu_time in
     let wall_span = Ptime.diff curr_wall_time t.last_wall_time in
     let wall_delta = Ptime.Span.to_float_s wall_span in
-    t.last_cpu_time <- curr_cpu_time;
-    t.last_wall_time <- curr_wall_time;
+    { last_cpu_time = curr_cpu_time; last_wall_time = curr_wall_time }
+    |> fun t ->
     if wall_delta <= 0.000001 then 0.0
     else
       let pct = cpu_delta /. wall_delta *. 100.0 in
--- a/autoscaler.ml
+++ b/autoscaler.ml
@@ -1,0 +1,81 @@
+module Cpu_monitor = struct
+  let rusage_to_float (sec, usec) =
+    let s = Int64.to_float sec in
+    let u = float_of_int usec in
+    s +. (u /. 1_000_000.0)
+
+  let get_total_cpu_time (r : Vmm_core.Stats.rusage) =
+    let user_t = rusage_to_float r.utime in
+    let sys_t = rusage_to_float r.stime in
+    user_t +. sys_t
+
+  type t = { mutable last_cpu_time : float; mutable last_wall_time : Ptime.t }
+
+  let create now (initial_rusage : Vmm_core.Stats.rusage) =
+    { last_cpu_time = get_total_cpu_time initial_rusage; last_wall_time = now }
+
+  let measure t now (current_rusage : Vmm_core.Stats.rusage) =
+    let curr_cpu_time = get_total_cpu_time current_rusage in
+    let curr_wall_time = now in
+    let cpu_delta = curr_cpu_time -. t.last_cpu_time in
+    let wall_span = Ptime.diff curr_wall_time t.last_wall_time in
+    let wall_delta = Ptime.Span.to_float_s wall_span in
+    t.last_cpu_time <- curr_cpu_time;
+    t.last_wall_time <- curr_wall_time;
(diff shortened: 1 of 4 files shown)


unikernel.ml
```diff
@@ -2733,6 +2740,132 @@ struct
...
+            m "Error finding cluster group for unikernel %s: %s" unikernel_name
+              err);
+        Ok ()
+    | Ok (group, key) -> (
+        match
+          Autoscaler.Cluster_manager.check_group_status group key now rusage
+        with
+        | Ok (Autoscaler.Normal usage_pct) ->
+            Logs.info (fun m ->
+                m "[%s] Normal load: %.2f%%" unikernel_name usage_pct);
```
hannesm: should this be on the debug log level?

autoscaler.ml
```diff
@@ -0,0 +1,232 @@
...
+
+type t = { mutable monitor : Cpu_monitor.t; mutable last_cpu_usage : float }
+
+type status =
+  | Overloaded of
+      t (* the usage of the clone or cm that is checked in the group *)
+  | Pending of int * float
+    (* We have a high load but waiting for required number of polls to complete*)
+  | Cooldown of float
+  | Normal of float
```
hannesm: I don't quite understand the different types used here and the reasoning behind it. `t` is only used in the Overloaded case, otherwise a float is used (or a float paired with an int)...

autoscaler.ml
```diff
@@ -0,0 +1,232 @@
...
+    mutable consecutive_high_ticks : int;
+  }
+
+  let clusters : (string, group) Hashtbl.t = Hashtbl.create 10
+
+  let in_cooldown now = function
+    | { last_scale_action = None; _ } -> false
+    | { last_scale_action = Some t; _ } ->
+        let span = Ptime.diff now t in
+        Ptime.Span.to_float_s span < cooldown_period
```
hannesm: not sure I understand what this is supposed to do, but there may be the case that we (re)start mollymawk, and it figures out that some unikernel was scaled up previously -- so we don't know the time when this happened, but mollymawk should be able to scale it down...

autoscaler.ml
```diff
@@ -0,0 +1,232 @@
...
+    let curr_wall_time = now in
+    let cpu_delta = curr_cpu_time -. t.last_cpu_time in
+    let wall_span = Ptime.diff curr_wall_time t.last_wall_time in
+    let wall_delta = Ptime.Span.to_float_s wall_span in
+    t.last_cpu_time <- curr_cpu_time;
+    t.last_wall_time <- curr_wall_time;
+    if wall_delta <= 0.000001 then 0.0
+    else
+      let pct = cpu_delta /. wall_delta *. 100.0 in
+      Float.min 100.0 (Float.max 0.0 pct)
```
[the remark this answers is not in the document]
PizieDust: Here, for above 100 you are correct in that multiple cpus can cause this. 
however for below 0, I try to imagine a scenario where a unikernel reboots and we try to compute it's new value againts a previous value which could potentially be negative? I  could be wrong about this

autoscaler.ml
PizieDust: the float is `cpu_usage` in the three cases and the int is the number of ticks we have checked for in case we have a high usage. we can make it uniform since we can get the `cpu_usage` from t

autoscaler.ml
```diff
@@ -0,0 +1,232 @@
...
+module Cluster_manager = struct
+  type group = {
+    primary : string * t;
+    mutable clones : (string * t) list;
+    mutable last_scale_action : Ptime.t option;
+    mutable next_id : int;
+    mutable consecutive_high_ticks : int;
+  }
+
+  let clusters : (string, group) Hashtbl.t = Hashtbl.create 10
```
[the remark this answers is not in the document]
PizieDust: Thanks for pointing this out.
I think if we don't receive stats for 15mins we can assume this vm is destroyed and prune it.

autoscaler.ml
PizieDust: normally the cooldown check is to tell us if we recently scaled up. the assumption is that if we just scaled up maybe the cpu_usage spikes haven't stablized yet and we should wait for the coodown duration to check if we need another scale action.

With the way we name the clones, if mollymawk restarts, it will be able to figure out what the clones are and either leave them in place, or scale them down (once we have scaling down implemented)

unikernel.ml
```diff
@@ -2733,6 +2740,132 @@ struct
...
+                        m
+                          "[%s] OVERLOAD CONFIRMED. Initiating Scale-Up! New \
+                           clone name: %s"
+                          unikernel_name new_name);
+                    Ok ()))
+        | Error err ->
+            Logs.err (fun m ->
+                m "Error checking cluster group status for unikernel %s: %s"
+                  unikernel_name err);
+            Ok ())
```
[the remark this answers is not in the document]
PizieDust: yes at the moment we don't have per-unikernel configuration. all unikernels can be scaled up if they have significant load.

hannesm:
Coming back to this, I suggest and ask:
- as proposed here, for each user, for each unikernel, a stats command is issued. But `` `Stats_cmd `Stats_subscribe `` will subscribe to the whole ARC of unikernels -> you will receive a stream of `` `Stats_data of Stats.t `` -- and will not need to do lots of communication to albatross (please see how e.g. console output is consumed, this is very much the same)
- is there code for disabling / enabling / having the scaling up/down threshold configured (per user / per unikernel)? If not, this is a blocker, since there are unikernels such as opam-mirror which do a lot of computation at startup (computing and verifying hashes), which we don't want to duplicate
- autoscaling of unikernels that require block devices: there should be somewhere a safeguard and this should be rejected
- `prune_dead_clusters` is a function you implemented, but I couldn't find it being called anywhere

PizieDust:
This PR is now split into multipe PRs:
- Autoscaling module only: https://github.com/robur-coop/mollymawk/pull/250
- Metrics handler: https://github.com/robur-coop/mollymawk/pull/251
- Scale down vms: https://github.com/robur-coop/mollymawk/pull/252
- Prune dead clusters: https://github.com/robur-coop/mollymawk/pull/253
- Spawn a new clone: https://github.com/robur-coop/mollymawk/pull/254

PR #222, opened 2026-03-21
Console output for deceased unikernels

### LINKED ISSUE #126: Show the last console output of a deceased unikernel
I've tried to deploy a unikernel but the argument i gave were not the right ones, which made the unikernel crash.
However no information about that is shown on the interface so the deployment seemed to have simply failed internally despite saying it succeeded.

It would be nice for example to have deceased unikernel shown on the dashboard as greyed-out or something like that, so we can click on their log if that feature is available.

PizieDust:
fixes https://github.com/robur-coop/mollymawk/issues/126

This PR checks which unikernels are deceased from a user and we can view the last logs of that unikernel.
We now have two tabs on the dashboard:
- one for live unikernels
- the other for deceased unikernels
image

On the deceased unikernels tab, we can see all unikernels deployed by the user which are not running:
image

and if we click on `view logs` we can see the console output of that deceased unikernel:
image

### DIFF
--- a/albatross.ml
+++ b/albatross.ml
@@ -586,6 +586,23 @@ module Make (S : Tcpip.Stack.V4V6) = struct
     | Ok (vmm_name, certificates) ->
         raw_query stack t ~name:vmm_name certificates cmd (console vmm_name f)
 
+  let query_deceased stack t ~domain ~name =
+    let cmd = `Console_cmd `Console_list_inactive in
+    match certs t domain name cmd with
+    | Error str -> Lwt.return (Error str)
+    | Ok (vmm_name, certificates) ->
+        raw_query stack t ~name:vmm_name certificates cmd (fun tls_flow d ->
+            let open Lwt.Infix in
+            TLS.close tls_flow >|= fun () ->
+            match decode_reply d with
+            | Error e -> Error e
+            | Ok (_hdr, `Success (`Consoles names)) -> Ok names
+            | Ok w ->
+                Error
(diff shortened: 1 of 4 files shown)


albatross.ml
```diff
@@ -586,6 +586,23 @@ module Make (S : Tcpip.Stack.V4V6) = struct
...
+            let open Lwt.Infix in
+            TLS.close tls_flow >|= fun () ->
+            match decode_reply d with
+            | Error e -> Error e
+            | Ok (_hdr, `Success (`Consoles names)) -> Ok names
+            | Ok w ->
+                Error
+                  (Fmt.str "unexpected reply to console list inactive: %a"
+                     (Vmm_commands.pp_wire ~verbose:false)
+                     w))
```
hannesm: why is this needed? what is the standard query missing?

(we need the query_console and query_block_dump due to the streaming interface)

hannesm:
code looks fine, I find the numerous css stuff and javascript code embedded as strings hard to read.

I'm wondering whether e.g. the css stuff could be factored out in a common module.

PizieDust:
> the UI looks great. thanks a lot. I'm curious why you use a tab for those unikernels? in my intuition I'd have used the same tab, but a slightly different colour (greyed out?).
> 
> I'll later look through the code.

The choice to use tabs was so it can be obvious at first glance where to find what...

hannesm:
thanks, I'll try to get the other VM types landed before re-deploying albatross & mollymawk

Issue #229, opened 2026-04-14
improve console output for deceased unikernels


hannesm:
#222 #126 I looked at this today, and would appreciate if I could be able to scroll in the window -- at the moment, I have a hard time to see the entire output... see the screenshot below, especially the line starting with `[2026-04-14T13:56:16-00:00] 2026-04-14T13:56:16-00:00: [INFO] [application]` is cut (and may have useful information in there).

There are ways to get the entire message, such as inspection of the content or selecting with the mouse and pasting the contents somewhere...

<img width="1582" height="891" alt="Image" src="https://github.com/user-attachments/assets/0c136210-5ad6-44e7-8f33-f948e95648de" />

PizieDust:
Thanks. for this, I am now thinking we can ditch the modal view. On builder-web I just saw a more preferable way to display such data, we can display it on a new browser tab and display the entire text

hannesm:
Another issue is, as seen in the screenshot, the output is reversed (newest output first).

PizieDust:
> Another issue is, as seen in the screenshot, the output is reversed (newest output first).

This is fixed in the newest commit:

<img width="2791" height="1684" alt="Image" src="https://github.com/user-attachments/assets/e9b67f34-69fa-49af-96d5-0bf0bca89f3c" />

hannesm:
thanks for your work in #230

PR #230, opened 2026-04-15
Display console output for deceased unikernels in a new tab

PizieDust:
This PR addresses https://github.com/robur-coop/mollymawk/issues/229


PizieDust:
image

hannesm:
I understand that this makes the issue I reported disappear. At the same time, the output is rather strange to read with all the json encoding, I'd expect as a unikernel operator to see the output as pre-formatted lines...

PizieDust:
now we have:

image

Issue #191, opened 2025-12-09
Simplify unikernel deployment


PizieDust:
Currently the deploy form tries to get as many parameters configured as possible. 
A better user experience will be to let them configure only what is necessary, and the rest of the configurations be put under an `Advanced Configuration` heading that will then let the user jump into more parameters. 

Secondly, if a user is only given 1 cpu_id to deploy on, there's no need allowing them a dropdown to select out of 1 option. This can be auto-selected, and this also applies to other resources i.e block devices, network interfaces. 

Memory should default to 32MB, unless the user used advanced configuration to modify it. 

cc @hannesm

PizieDust:
closed by PR #233

PR #232, opened 2026-04-17
Simpler deployment ui

PizieDust:
This PR tackles some of the issues mentioned in https://github.com/robur-coop/mollymawk/issues/191

When deploying, the simpler UI contains:
- the name input for the unikernel
- the type of unikernel (solo5, bhyve)
- the unikernel arguments
- an input to upload the unikernel binary

The advanced configuration button displays more settings:
- cpu ids
- memory
- startup priority
- network devices
- block devices
- fail behaviour
- force create toggle

### Simple deployment form
image

### Advanced configuration
image

### DIFF
--- a/unikernel_create.ml
+++ b/unikernel_create.ml
@@ -144,7 +144,11 @@ let unikernel_create_layout ~(user_policy : Vmm_core.Policy.t) unikernels
   let memory_left = user_policy.memory - total_memory_used in
   Tyxml_html.(
     section
-      ~a:[ a_class [ "col-span-7 p-4 bg-gray-50 my-1" ] ]
+      ~a:
+        [
+          a_class [ "col-span-7 p-4 bg-gray-50 my-1" ];
+          Unsafe.string_attrib "x-data" "{ advanced: false }";
+        ]
       [
         div
           ~a:[ a_class [ "px-3 flex justify-between items-center" ] ]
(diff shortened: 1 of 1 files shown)


hannesm:
I don't have much insight how this is supposed to work, from the screenshots I've some questions:
- what happens if the unikernel needs a network device?
- what about block devices?
- it seems that CPUids are automatically selected? Is this only a single one? Are network/block devices as well be automatically selected?

Will we then always have to click on "advanced options"?

Is a simple deployment UI also for BHyve? Or should it focus on unikernels?


Maybe it would help to describe a bit in text what the workflow for a unikernel (with, with multiple, without) network & block devices should look like (or what is possible with the current setup, what kind of auto-completion should be done (vs what not to do))? Maybe some discussion would help?

PizieDust:
> * what happens if the unikernel needs a network device?
> * what about block devices?

The user has to click on "Advanced configuration" to set this up. both for network devices and block devices.


> * it seems that CPUids are automatically selected? Is this only a single one? Are network/block devices as well be automatically selected?

For CPUIds we try to select the one with the least number of unikernels deployed to it. 
Given that mollymawk doesn't know what the unikernel needs in terms of network/block devices, it can't automatically select them.

> Will we then always have to click on "advanced options"?

At the moment, yes this is the case.

> Is a simple deployment UI also for BHyve? Or should it focus on unikernels?

This is also the case for BHyve

> Maybe it would help to describe a bit in text what the workflow for a unikernel (with, with multiple, without) network & block devices should look like (or what is possible with the current setup, what kind of auto-completion should be done (vs what not to do))? Maybe some discussion would help?

So basically, we have a name, unikernel type, binary upload and unikernel arguments. Everything else is just hidden in the Advanced Configuration menu.
For a unikernel like hello, we don't need anything, just a name and the binary and this can deploy without us looking at the Advanced Configurations. While this is as simple as it gets, for other unikernel vms which some require a block device or a network device, mollymawk doesn't yet know how to get the manifest to figure out which devices are required so we can't automate this part of the deployment yet.

unikernel_create.ml
```diff
@@ -197,173 +206,241 @@ let unikernel_create_layout ~(user_policy : Vmm_core.Policy.t) unikernels
...
+                hr ();
+                div
+                  [
+                    label
+                      ~a:[ a_class [ "block font-medium" ] ]
+                      [ txt "Network Interfaces" ];
+                    Utils.dynamic_dropdown_form
+                      (Vmm_core.String_set.elements user_policy.bridges)
+                      ~get_label:(fun bridge -> bridge)
+                      ~get_value:(fun bridge -> bridge)
```
reynir: This is perfectly fine. Can be rewritten using `Fun.id`
```suggestion
                      ~get_label:Fun.id
                      ~get_value:Fun.id
```

unikernel_create.ml
```diff
@@ -197,173 +206,241 @@ let unikernel_create_layout ~(user_policy : Vmm_core.Policy.t) unikernels
...
+                  [
+                    cpu_multiselect cpu_usage_count;
+                    Utils.increment_or_decrement_ui ~id:"unikernel-memory"
+                      ~max_value:memory_left ~min_value:0 ~default_value:32
+                      ~figure_unit:"MB" ~step:32 ~label':"Memory" ();
+                    Utils.increment_or_decrement_ui ~id:"startup-priority"
+                      ~max_value:100 ~min_value:0 ~default_value:50
+                      ~label':"Startup Priority" ();
+                  ];
+                (* BHyve Specific Options *)
```
reynir, line 365: I'm not up to date on what interface albatross offers for BHyve.

unikernel_create.ml
```diff
@@ -5,9 +5,14 @@ let input_classes =
    ring-primary-200 focus:ring-primary-200 focus:ring-[1px] focus:outline-none"
 
 let cpu_multiselect cpu_usage_count =
+  let sorted_cpus =
+    cpu_usage_count |> List.sort (fun (_, c1) (_, c2) -> Int.compare c1 c2)
+  in
+  let default_cpu =
+    match sorted_cpus with (id, _) :: _ -> string_of_int id | [] -> ""
+  in
```
reynir, line 13: Maybe a better default is to allow all CPUs? Somewhat recently we made it a list of CPUs that we are *allowed* to execute on instead of *the* CPU to execute on. If we default to the first CPU that core might get crowded quickly. Does Mollymawk allow multiple CPUs?

unikernel_create.ml
```diff
@@ -197,173 +206,241 @@ let unikernel_create_layout ~(user_policy : Vmm_core.Policy.t) unikernels
...
+                  ~a:[ a_class [ "grid grid-cols-3 gap-3" ] ]
+                  [
+                    cpu_multiselect cpu_usage_count;
+                    Utils.increment_or_decrement_ui ~id:"unikernel-memory"
+                      ~max_value:memory_left ~min_value:0 ~default_value:32
+                      ~figure_unit:"MB" ~step:32 ~label':"Memory" ();
+                    Utils.increment_or_decrement_ui ~id:"startup-priority"
+                      ~max_value:100 ~min_value:0 ~default_value:50
+                      ~label':"Startup Priority" ();
+                  ];
```
reynir, line 364: Should we have a brief explanation on what startup priority means?

reynir (commented):
I wonder with this new flow if we try to deploy a unikernel that requires one network interface how that works. The default is to have no network interfaces configured, so deploying that will lead to a failure from solo5-hvt/albatross. I think almost all real-world unikernels use at least one device (usually network) and so we would almost always go into the advanced options (are they then advanced?).

Maybe with the discussed change to use sane defaults for network interfaces this would work - but then that requires that we have the manifest, and at the moment we can only have that if we deploy via builder-web.

In JavaScript would it be possible to parse a file selected in a form input to be uploaded and get the manifest? That would likely be a bigger work item. So maybe an easier path for now is to move network and block devices back out from advanced settings, and then we can look into moving them back under advanced if we can synthesize sane defaults.

hannesm:
I agree with @reynir. and think that deploying via builds.robur.coop will usually be the default (so we have a manifest) -- but we should think about how to get the manifest from a unikernel image. According to https://developer.mozilla.org/en-US/docs/Web/API/File_API/Using_files_from_web_applications there are ways to read such a file... Since we have ocaml-solo5-elftool, we could compile this to JavaScript to read and extract the manifest on the client side... But I guess we should discuss this in a separate issue, and not block this PR.

I especially agree that any unikernel that is interesting (and not hello world) will need a network interface - so removing this from the simple deployment will require everyone deploying an interesting unikernel to use the advanced deployment UI (and then this simple one won't help for many deployments).

hannesm:
Also, as mentioned yesterday, from my perspective the simple UI for network interfaces will be that these are preselected -- but a user still can modify such a selection and modify network interfaces.

unikernel_create.ml
PizieDust: yes mollymawk allows multiple CPUs, but the idea here is that the cpu with the least number of unikernels should be ranked first

unikernel_create.ml
PizieDust: yes we can do that

PizieDust:
The current ci failure is unrelated to this PR

PizieDust:
I'll go ahead and merge this. The further issue about getting manifest devices when the binary is uploaded locally the issue is opened at https://github.com/robur-coop/mollymawk/issues/238

PR #233, opened 2026-05-01
Deploy vms directly from builds.robur.coop

PizieDust:
This pr introduces direct deployment from builds.robur.coop. We use HTMX to perform background API calls and populate the DOM without refreshing the page.

Basically when the deployment form is opened:
- mollymawk attempts to query builds.robur.coop to get a list of available unikernels. This list is populated in a dropdown select on the deployment form.
- the user selects which vm (clicking on the job name) they want to deploy, mollymawk (via htmx) queries builds.robur.coop a second time to get it's manifest (which block devices and network devices are needed). this is then used to populated the form sections handling block devices and network devices.
- when the user clicks on deploy, mollymawk uses the job name to stream the latest binary from builds.robur.coop to deploy the unikernel.

### Deployment modes
We have two deployment modes:
- Robur builder: the default mode, which lets you select any unikernel from builds.robur.coop
- Manual mode: which lets you upload the unikernel binary manually on the form
image

### VMs available
A dropdown with all available reproducible builds (only unikernels) from builds.robur.coop. Selecting any will query the specific vm to see what devices it nees

image

### Configuring devices
Once the manifest devices are known, the advanced config panel opens and autofills this information for any network interfaces needed or block devices needed. Manual  intervention is needed to map the names to the correct hosts.

image

### DIFF
--- a/assets/main.js
+++ b/assets/main.js
@@ -1362,3 +1362,38 @@ function areCharactersValid(s) {
 	}
 	return true;
 }
+
+window.mapFields = function (field_id, detail, options) {
+	if (!detail || !detail[field_id]) return [];
+	let reqs = detail[field_id];
+	if (field_id === 'network') {
+		let unassigned = [...options];
+		let fields = reqs.map(n => {
+			let idx = unassigned.findIndex(o => o.value === n || o.label === n);
+			if (idx !== -1) {
+				let match = unassigned.splice(idx, 1)[0];
+				return { title: n, selected: match.value };
+			}
+			return { title: n, selected: '' };
+		});
+		if (reqs.length === 1 && options.length === 1 && fields[0].selected === '') {
+			fields[0].selected = options[0].value;
+		} else if (reqs.length === options.length && unassigned.length === 1) {
+			let unmatched = fields.find(f => f.selected === '');
+			if (unmatched) unmatched.selected = unassigned[0].value;
+		}
+		return fields;
+	} else if (field_id === 'block') {
+		let job_name = detail.job_name;
+		return reqs.map(n => {
+			let exact = options.find(o => o.value === n);
+			if (exact) return { title: n, selected: exact.value };
+			if (job_name) {
+				let job_match = options.find(o => o.value === job_name || o.value === job_name + '-data');
+				if (job_match) return { title: n, selected: job_match.value };
+			}
+			return { title: n, selected: '' };
+		});
+	}
+	return reqs.map(n => ({ title: n, selected: '' }));
+};
--- a/assets/main.js
+++ b/assets/main.js
@@ -463,31 +463,27 @@ async function deployUnikernel(albatross_instance) {
 		if (!networkData || !blockData) {
 			return;
 		}
-		if (deploy_mode !== 'builder') {
-			if (typ === 'solo5' && !binary) {
-				showError(formAlert, "Please upload the unikernel image");
-				return;
-			}
+		if (deploy_mode === 'manual' && !binary) {
+			showError(formAlert, "Please upload the unikernel image");
+			return;
 		}
 
 		buttonLoading(deployButton, true, "Deploying...");
 
 		let specificConfig = {};
 
-		if (deploy_mode !== 'builder') {
-			if (typ === 'solo5') {
-				const argumentsString = document.getElementById("unikernel-arguments").value.trim();
-				specificConfig = {
-					arguments: argumentsString ? argumentsString.split(/\s+/) : []
-				};
-			} else if (typ === 'bhyve') {
-				const vcpus = document.getElementById("numcpus").innerText;
-				const bootPartition = document.getElementById("linux-boot-partition").value;
-				specificConfig = {
-					vcpus: Number(vcpus),
-					linux_boot_partition: bootPartition === "" ? null : bootPartition
-				};
-			}
+		if (typ === 'solo5') {
+			const argumentsString = document.getElementById("unikernel-arguments").value.trim();
+			specificConfig = {
+				arguments: argumentsString ? argumentsString.split(/\s+/) : []
+			};
+		} else if (typ === 'bhyve') {
+			const vcpus = document.getElementById("numcpus").innerText;
+			const bootPartition = document.getElementById("linux-boot-partition").value;
+			specificConfig = {
+				vcpus: Number(vcpus),
+				linux_boot_partition: bootPartition === "" ? null : bootPartition
+			};
 		}
 
 		const unikernel_config = {
--- a/assets/main.js
+++ b/assets/main.js
@@ -434,72 +434,89 @@ function gatherFieldsForDevices(fieldId, targetKey, formAlert) {
 async function deployUnikernel(albatross_instance) {
 	const formAlert = document.getElementById("form-alert");
 	const deployButton = document.getElementById("deploy-button");
-	const name = document.getElementById("unikernel-name").value;
-	const type = document.getElementById("unikernel-type").value;
-	const startup = document.getElementById("startup-priority").innerText;
-	const fail_behaviour = document.getElementById("restart-on-fail").checked;
-	const force_create = document.getElementById("force-create").checked;
-	const memory = document.getElementById("unikernel-memory").innerText;
-	const molly_csrf = document.getElementById("molly-csrf").value;
-	const cpuidsStr = document.getElementById("cpuids").value;
-	const cpuids = cpuidsStr ? cpuidsStr.split(',').map(Number) : [];
-	const networkData = gatherFieldsForDevices("network", "network_interfaces", formAlert);
-	const blockData = gatherFieldsForDevices("block", "block_devices", formAlert);
-	if (networkData === null || blockData === null) {
+	const handleFailure = (msg) => {
+		postAlert("bg-secondary-300", msg);
+		showError(formAlert, msg);
 		buttonLoading(deployButton, false, "Deploy");
-		return;
-	}
-	if (!isValidName(name)) {
-		showError(formAlert, "Please provide a valid name (alphanumeric, no spaces or special symbols, must not start with a hyphen, max 64 chars).");
-		document.getElementById("unikernel-name").classList.add("border-secondary-500", "ring-secondary-500");
-		buttonLoading(deployButton, false, "Deploy");
-		return;
-	}
-	let specificConfig = {};
-	const binary = document.getElementById("unikernel-binary").files[0];
-	if (type === 'solo5') {
-		if (!binary) {
-			showError(formAlert, "Please upload the unikernel image");
-			buttonLoading(deployButton, false, "Deploy");
+	};
+	try {
+		const name = document.getElementById("unikernel-name").value;
+		const typ = document.getElementById("unikernel-type").value;
+		const startup = document.getElementById("startup-priority").innerText;
+		const fail_behaviour = document.getElementById("restart-on-fail").checked;
+		const force_create = document.getElementById("force-create").checked;
+		const memory = document.getElementById("unikernel-memory").innerText;
+		const molly_csrf = document.getElementById("molly-csrf").value;
+		const cpuidsStr = document.getElementById("cpuids").value;
+		const deployModeNode = document.querySelector('input[name="deploy_mode"]:checked');
+		const deploy_mode = deployModeNode ? deployModeNode.value : 'manual';
+		const binaryInput = document.getElementById("unikernel-binary");
+		const binary = (binaryInput && binaryInput.files) ? binaryInput.files[0] : null;
+		if (!isValidName(name)) {
+			document.getElementById("unikernel-name").classList.add("border-secondary-500", "ring-secondary-500");
+			showError(formAlert, "Please provide a valid name (alphanumeric, no spaces or special symbols, must not start with a hyphen, max 64 chars).");
 			return;
 		}
-		const argumentsString = document.getElementById("unikernel-arguments").value.trim();
-		specificConfig = {
-			typ: "solo5",
-			arguments: argumentsString ? argumentsString.split(/\s+/) : []
+
+		const networkData = gatherFieldsForDevices("network", "network_interfaces", formAlert);
+		const blockData = gatherFieldsForDevices("block", "block_devices", formAlert);
+		if (!networkData || !blockData) {
+			return;
+		}
+		if (deploy_mode !== 'builder') {
+			if (typ === 'solo5' && !binary) {
+				showError(formAlert, "Please upload the unikernel image");
+				return;
+			}
+		}
+
+		buttonLoading(deployButton, true, "Deploying...");
+
+		let specificConfig = {};
+
+		if (deploy_mode !== 'builder') {
+			if (typ === 'solo5') {
+				const argumentsString = document.getElementById("unikernel-arguments").value.trim();
+				specificConfig = {
+					arguments: argumentsString ? argumentsString.split(/\s+/) : []
+				};
+			} else if (typ === 'bhyve') {
+				const vcpus = document.getElementById("numcpus").innerText;
+				const bootPartition = document.getElementById("linux-boot-partition").value;
+				specificConfig = {
+					vcpus: Number(vcpus),
+					linux_boot_partition: bootPartition === "" ? null : bootPartition
+				};
+			}
+		}
+
+		const unikernel_config = {
+			typ: typ,
+			cpuids: cpuidsStr ? cpuidsStr.split(',').map(Number) : [],
+			startup: Number(startup),
+			fail_behaviour: fail_behaviour ? { "restart": fail_behaviour } : "quit",
+			memory: Number(memory),
+			...networkData,
+			...blockData,
+			...specificConfig
 		};
-	} else if (type === 'bhyve') {
-		const vcpus = document.getElementById("numcpus").innerText;
-		const bootPartition = document.getElementById("linux-boot-partition").value;
-		specificConfig = {
-			typ: "bhyve",
-			vcpus: Number(vcpus),
-			linux_boot_partition: bootPartition == "" ? null : bootPartition
-		};
-	}
-	buttonLoading(deployButton, true, "Deploying...");
-	let formData = new FormData();
-	const unikernel_config = {
-		cpuids: cpuids,
-		startup: Number(startup),
-		fail_behaviour: fail_behaviour ? { "restart": fail_behaviour } : "quit",
-		memory: Number(memory),
-		...networkData,
-		...blockData,
-		...specificConfig
-	};
-	formData.append("albatross_instance", albatross_instance);
-	formData.append("unikernel_name", name);
-	formData.append("typ", type);
-	formData.append("unikernel_config", JSON.stringify(unikernel_config));
-	formData.append("unikernel_force_create", force_create);
-	formData.append("molly_csrf", molly_csrf);
-	if (binary) {
-		formData.append("binary", binary);
-	} else {
-		formData.append("binary", "");
-	}
-	try {
+
+		const formData = new FormData();
+		formData.append("albatross_instance", albatross_instance);
+		formData.append("unikernel_name", name);
+		formData.append("unikernel_config", JSON.stringify(unikernel_config));
+		formData.append("unikernel_force_create", force_create);
+		formData.append("molly_csrf", molly_csrf);
+		formData.append("deploy_mode", deploy_mode);
+
+		if (deploy_mode === 'builder') {
+			const builderJobSelect = document.getElementById("builder-job-select");
+			if (builderJobSelect) formData.append("builder_job", builderJobSelect.value);
+			formData.append("binary", "");
+		} else {
+			formData.append("binary", binary);
+		}
+
 		const response = await fetch("/api/unikernel/create", {
 			method: 'POST',
 			body: formData
@@ -509,19 +526,14 @@ async function deployUnikernel(albatross_instance) {
 		if (data.status === 200 && data.success) {
 			showError(formAlert, "Successfully updated", true);
 			postAlert("bg-primary-300", `${name} has been deployed successfully.`);
-			setTimeout(function () {
+			setTimeout(() => {
 				window.location.href = "/dashboard";
 			}, 2000);
 		} else {
-			handleFailure(data.data);
+			handleFailure(data.data || "An error occurred during deployment.");
 		}
 	} catch (error) {
-		handleFailure(error);
-	}
-	function handleFailure(msg) {
-		postAlert("bg-secondary-300", msg);
-		showError(formAlert, msg);
-		buttonLoading(deployButton, false, "Deploy");
+		handleFailure(error.message || "A network error occurred.");
 	}
 }
 
--- a/assets/main.js
+++ b/assets/main.js
@@ -1365,7 +1365,7 @@ window.mapFields = function (field_id, detail, options) {
 	if (field_id === 'network') {
 		let unassigned = [...options];
 		let fields = reqs.map(n => {
-			let idx = unassigned.findIndex(o => o.value === n || o.label === n);
+			let idx = unassigned.findIndex(o => o.value === n);
 			if (idx !== -1) {
 				let match = unassigned.splice(idx, 1)[0];
 				return { title: n, selected: match.value };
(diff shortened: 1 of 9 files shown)


builder_web.ml
```diff
@@ -302,3 +302,119 @@ let compare_of_json = function
...
+  | Ok (`Assoc xs) -> (
+      match (Utils.Json.get "uuid" xs, Utils.Json.get "solo5_manifest" xs) with
+      | Some (`String uuid), Some (`Assoc manifest) -> (
+          match Utils.Json.get "devices" manifest with
+          | Some (`List devices) ->
+              let req_devices =
+                List.filter_map
+                  (fun d -> Result.to_option (device_of_json d))
+                  devices
+              in
```
hannesm: What about the failure semantics here? It seems you try to decode the manifest, and if an entry fails, you skip it.

This means that we proceed with deployment even though we knew a device is missing (which couldn't be parsed).

I propose to fail in such a case. I may be wrong since I can't follow the code entirely.

builder_web.ml
```diff
@@ -302,3 +302,119 @@ let compare_of_json = function
...
+              Ok { uuid; devices = req_devices }
+          | Some _ -> Error (`Msg "solo5_manifest.devices is not a list")
+          | None -> Ok { uuid; devices = [] } (* No devices required *))
+      | Some (`String uuid), None -> Ok { uuid; devices = [] }
```
hannesm: I'd appreciate to understand this -- so there was no manifest associated with this unikernel. What should the failure semantics be?

unikernel.ml
```diff
@@ -3183,7 +3201,17 @@ struct
             check_meth `GET (fun () ->
                 authenticate store reqd
                   (albatross_instance "/unikernel/deploy"
-                     (deploy_form stack store)))
+                     (deploy_form stack store http_client)))
+        | "/api/deploy/manifest" ->
+            check_meth `GET (fun () ->
+                match get_query_parameter "job" with
```
hannesm: I'm confused now, why is there this new endpoint with a `job` parameter?

PizieDust: This endpoint uses the job (name of the vm on builds.robur.coop) to then query builds.robur.coop for the manifest of the vm.
So, when we first load the page, we have a dropdown of unikernels from builds.robur.coop.
When we select any on the dropdown, all what we have is the job name, and with this job name we then send another request to builds.robur.coop to get it's manifest devices.
so the endpoint `/api/deploy/manifest` is what we call after selecting a vm on the dropdown, and calling this endpoint will trigger a background request to builds.robur.coop to get us the manifest

PizieDust: so we send three requests to builds.robur.coop
- first request: to get all unikernels from builds.robur.coop
- second request: to get the manifest devices of a specific vm
- third request: to stream the binary of the vm

we can reduce the number of request to two if the first request already includes which devices the vms need in addition to providing the names of the vms.

@reynir what do you think about this?

hannesm:
Thanks for this PR, I have some questions from the other PR (https://git.robur.coop/robur/builder-web/pulls/40#issuecomment-2690) in builder-web:
- if there would be a json output that includes "name", "description" (synopsis), "target" (hvt / package / xen), and "required network & block devices" -- would this help mollymawk a lot?
- From what I think, on the mollymawk side we could then easily filter all non-hvt builds, and for a user that has access to N network interfaces only list all unikernels that need <= N network devices (same for block devices, if the user doesn't have access to block devices, we can filter all unikernels requiring block devices).

And in addition, I have some more notes:
- this PR, if I understand, has a dropdown of all unikernels from builds.robur.coop!? When is this list populated (on each unikernel create form? once per mollymawk session?)
- I still wonder about the "default memory" for a unikernel -- what is its value? Should that be something provided as well on a per-unikernel basis (it sounds to me, a hello world unikernel will be able to run with a very small amount of memory, while unipi will need quite a bit more).

Does this make sense to you?

PizieDust:
> * this PR, if I understand, has a dropdown of all unikernels from builds.robur.coop!? When is this list populated (on each unikernel create form? once per mollymawk session?)

this is done on each unikernel create form

> * I still wonder about the "default memory" for a unikernel -- what is its value? Should that be something provided as well on a per-unikernel basis (it sounds to me, a hello world unikernel will be able to run with a very small amount of memory, while unipi will need quite a bit more).

but how do we know the memory requirements of a unikernel? if this is something builds.robur.coop can provided it will be much appreciated. currently we default to 32MB

PizieDust:
> Thanks for this PR, I have some questions from the other PR (https://git.robur.coop/robur/builder-web/pulls/40#issuecomment-2690) in builder-web:
> 
> * if there would be a json output that includes "name", "description" (synopsis), "target" (hvt / package / xen), and "required network & block devices" -- would this help mollymawk a lot?
> * From what I think, on the mollymawk side we could then easily filter all non-hvt builds, and for a user that has access to N network interfaces only list all unikernels that need <= N network devices (same for block devices, if the user doesn't have access to block devices, we can filter all unikernels requiring block devices).

Yes this will be a very ideal case, if we have a json output which already provides all this information, then Mollymawk can get a lot smarter and the UX will be better too :)

unikernel.ml
reynir: Sounds good to me. We can either modify the existing endpoint to provide all this or add another one.

hannesm:
> Matching of network interfaces

As I can see in [afbec76](https://github.com/robur-coop/mollymawk/pull/233/commits/afbec7600384c7397dbefa782db13e0c6e7ad3ec), there's `o.value === n || o.label === n` -- I'm not sure what value and label are, but I think we should be one step smarter:
- if the user has access to a single network interface, and the unikernel requires exactly one, use it. This may be that the unikernel requires "service", but the user has access to "as250" or "private"
- if there are N network devices, and N - 1 can be easily mapped, assign the remaining one as well. As example, if the unikernel requires "service" and "management", and the user has access to "management" and "private", it is pretty likely that "service" will be mapped with "private" and "management" with "management".

For block devices, there are two scenarios (as discussed last week):
- this is a new unikernel, and no block device exists for it (yet) -- the user needs to create a new one
- the unikernel is named X, and there exists a block device X (or X-data) -- we can then map it easily. as example, we have a unikernel named opam-mirror and a block device named opam-mirror (although in the unikernel the name is "tar").

builder_web.ml
```diff
@@ -302,3 +302,119 @@ let compare_of_json = function
...
+      | Some (`Assoc sections) ->
+          let u1 =
+            match Utils.Json.get "Unikernels" sections with
+            | Some (`List l) -> l
+            | _ -> []
+          in
+          let u2 =
+            match
+              Utils.Json.get "Unikernels (with metrics reported to Influx)"
+                sections
```
reynir: For another time: maybe we should consider using a more machine readable key here in the builder-web json instead of the title. It seems brittle to me, and makes translations harder.

builder_web.ml
reynir: We get the unikernels listed in the "Unikernels" and "Unikernels (with metrics reported to Influx)" sections so it's reasonable to expect a manifest. However, there's not a strong coupling between these two parts of the code, so future changes could potentially include non-unikernel builds if we are not careful.

Maybe logging a warning is fine.

builder_web.ml
```diff
@@ -302,3 +302,119 @@ let compare_of_json = function
...
+            with
+            | Some (`List l) -> l
+            | _ -> []
+          in
+          let combine = u1 @ u2 in
+          let jobs =
+            List.filter_map (function `String s -> Some s | _ -> None) combine
+          in
```
reynir: Likely we want to change the response to be an object. So I think we should at least log a warning if the shape isn't as we expect.

unikernel.ml
```diff
@@ -1152,8 +1152,22 @@ struct
                   (Utils.Json.to_string (`Assoc json_dict))))
           `Bad_request
 
-  let deploy_form stack store albatross _ (user : User_model.user) reqd =
+  let deploy_manifest stack store http_client job albatross _
+      (user : User_model.user) reqd =
+    Builder_web.fetch_latest_build_manifest http_client job >>= function
+    | Ok manifest ->
+        user_blocks_by_instance stack albatross user.name >>= fun blocks ->
```
reynir: I think this value  `blocks` isn't used

unikernel.ml
```diff
@@ -1889,55 +1875,88 @@ struct
...
-                                      ("Albatross Response Error: " ^ err_str)
-                                      `Internal_server_error
-                                | Error (`Msg err_msg) ->
-                                    generate_http_error_response
-                                      ("Albatross JSON Error: " ^ err_msg)
-                                      `Internal_server_error))
+                            let is_builder =
+                              Option.value ~default:"manual" !deploy_mode_ref
+                              = "builder"
+                            in
```
reynir: We have that bound as `deploy_mode` so no need to use `Option.value`.

unikernel.ml
```diff
@@ -1889,55 +1875,88 @@ struct
...
+                                    http_client ~job ~version:"latest"
+                                    push_chunks
+                                    (execute_albatross_cmd push)
+                                  >>= fun (_, res) -> Lwt.return res
+                              | _ ->
+                                  generate_http_error_response
+                                    "Missing builder job or uuid" `Bad_request
+                            else
+                              execute_albatross_cmd (fun () ->
+                                  Lwt_stream.get contents)
```
reynir: It seems that in order to deploy from builder-web we still have to pass a value as `binary` in the multipart form. Is this intentional?

unikernel_create.ml
```diff
@@ -137,17 +137,43 @@ let cpu_multiselect cpu_usage_count =
...
+  in
+  let blocks_json =
+    `List (extract_names "BLOCK_BASIC") |> Yojson.Basic.to_string
+  in
+  Tyxml_html.(
+    script
+      (Unsafe.data
+         (Printf.sprintf
+            "window.dispatchEvent(new CustomEvent('populate-manifest', \
+             {detail: {network: %s, block: %s}}));"
```
reynir: The whole object could be one yojson earlier. This is fine though.

unikernel_create.ml
```diff
@@ -137,17 +137,43 @@ let cpu_multiselect cpu_usage_count =
...
+          Some (`String d.Builder_web.name)
+        else None)
+      manifest.devices
+  in
+  let networks_json =
+    `List (extract_names "NET_BASIC") |> Yojson.Basic.to_string
+  in
+  let blocks_json =
+    `List (extract_names "BLOCK_BASIC") |> Yojson.Basic.to_string
+  in
```
reynir: We don't do anything if we find unknown device types. The solo5 manifest has been incredibly stable, but it would be nice to handle unknown devices types anyway.

reynir: So what can we do? I think emitting a warning could be nice. "There's a device in the manifest with an unknown type. Deploying might fail" or so.

reynir: This can be done later in a separate PR (please open an issue if you don't implement it so we can track it).

unikernel_create.ml
```diff
@@ -162,6 +188,74 @@ let unikernel_create_layout ~(user_policy : Vmm_core.Policy.t) unikernels
...
+                      a_class
+                        [
+                          "flex flex-col items-center justify-center p-6 \
+                           border-2 rounded-lg cursor-pointer \
+                           transition-colors duration-200";
+                        ];
+                      Unsafe.string_attrib ":class"
+                        "deploy_mode === 'builder' ? 'border-primary-500 \
+                         bg-primary-100 text-primary-700' : 'border-gray-200 \
+                         hover:bg-gray-50 bg-white'";
```
reynir: I don't understand this bit.

unikernel_create.ml
```diff
@@ -162,6 +188,74 @@ let unikernel_create_layout ~(user_policy : Vmm_core.Policy.t) unikernels
...
+                    ]
+                  [
+                    input
+                      ~a:
+                        [
+                          a_class [ "sr-only" ];
+                          a_input_type `Radio;
+                          a_name "deploy_mode";
+                          a_value "builder";
+                          Unsafe.string_attrib "x-model" "deploy_mode";
```
reynir: Not sure about this either.

unikernel_create.ml
```diff
@@ -162,6 +188,74 @@ let unikernel_create_layout ~(user_policy : Vmm_core.Policy.t) unikernels
...
+                    input
+                      ~a:
+                        [
+                          a_class [ "sr-only" ];
+                          a_input_type `Radio;
+                          a_name "deploy_mode";
+                          a_value "manual";
+                          Unsafe.string_attrib "x-model" "deploy_mode";
+                        ]
+                      ();
```
reynir: Do we have two `deploy_mode` radio buttons with only one value? I don't quite understand this HTML.

unikernel_create.ml
```diff
@@ -190,13 +284,23 @@ let unikernel_create_layout ~(user_policy : Vmm_core.Policy.t) unikernels
...
                       ~a:
                         [
                           a_id "unikernel-type";
                           a_name "typ";
                           a_class [ input_classes ];
-                          a_onchange "toggleType()";
+                          Unsafe.string_attrib ":disabled"
+                            "deploy_mode === 'builder'";
+                          Unsafe.string_attrib "x-on:change"
+                            "solo5_options = ($event.target.value === 'solo5')";
```
reynir: Not sure about this either.

reynir (commented):
I didn't look too much into the javascript code or the html.

builder_web.ml
reynir: Also, it's already out of date with the main branch in builder-web: it's an object `{ name: "foo", synopsis: "this is the foo binary." }`. builds.robur.coop is still running an older version of builder-web however.

builder_web.ml
hannesm: I'm not sure why is there such json sections? IMHO this is useful for the HTML front page, but the json should not have it.

reynir: So the response has two fields: `jobs` and `jobs_by_section`. with the recent changes in builder-web (with solo5 manifest) I think we should use the `jobs` field and filter on the `latest` field being present **and** there is a solo5 manifest in there.

assets/main.js
```diff
@@ -1350,3 +1358,38 @@ function areCharactersValid(s) {
...
 	return true;
 }
+
+window.mapFields = function (field_id, detail, options) {
+	if (!detail || !detail[field_id]) return [];
+	let reqs = detail[field_id];
+	if (field_id === 'network') {
+		let unassigned = [...options];
+		let fields = reqs.map(n => {
+			let idx = unassigned.findIndex(o => o.value === n || o.label === n);
```
hannesm: I'm still unsure what the `value` and the `label` are, and what the difference is.

builder_web.ml
```diff
@@ -302,3 +302,126 @@ let compare_of_json = function
         (`Msg
            ("invalid json for builder_web diff, expected a dict: "
           ^ Utils.Json.to_string js))
+
+type device = { name : string; type_ : string }
```
hannesm: With what we support right now, I'd instead use

```OCaml
type device =
  | Network of string
  | Block of string
```

and then in e.g. `device_of_json` check for the right type (NET/BLOCK_BASIC).

builder_web.ml
```diff
@@ -302,3 +302,126 @@ let compare_of_json = function
...
+      | None -> Ok { uuid; devices = [] } (* No devices required *))
+  | Some (`String uuid), None -> Ok { uuid; devices = [] }
+  | _ -> Error (`Msg "Missing uuid or solo5_manifest")
+
+type job = { name : string; synopsis : string; manifest : device_manifest }
+
+let job_of_json = function
+  | `Assoc xs -> (
+      let name =
+        match Utils.Json.get "name" xs with Some (`String s) -> s | _ -> ""
```
hannesm: Can we use result here, and have an `Ok s` and an `Error` if the `"name"` is not a string?

builder_web.ml
```diff
@@ -302,3 +302,126 @@ let compare_of_json = function
...
+      in
+      let synopsis =
+        match Utils.Json.get "synopsis" xs with
+        | Some (`String s) -> s
+        | _ -> ""
```
hannesm: same as above

builder_web.ml
```diff
@@ -302,3 +302,126 @@ let compare_of_json = function
...
+          match Utils.Json.get "solo5_abi" latest with
+          | Some (`Assoc abi) -> (
+              match Utils.Json.get "target" abi with
+              | Some (`String "hvt") -> (
+                  match manifest_of_json (`Assoc latest) with
+                  | Ok manifest -> Ok (Some { name; synopsis; manifest })
+                  | Error e -> Error e)
+              | _ -> Ok None)
+          | _ -> Ok None)
+      | _ -> Ok None)
```
hannesm: error handling: I'd appreciate if we can be a bit louder if we receive unexpected json, and not have a catch-all and report None... the very least should be a log message so that the mollymawk operator can find a clue why something isn't listed.

builder_web.ml
```diff
@@ -302,3 +302,126 @@ let compare_of_json = function
...
+          match jobs_of_json json with
+          | Ok jobs -> Lwt.return jobs
+          | Error (`Msg e) ->
+              Logs.err (fun m -> m "Failed to fetch jobs: %s" e);
+              Lwt.return [])
+      | Error (`Msg e) ->
+          Logs.err (fun m -> m "Failed to fetch jobs: %s" e);
+          Lwt.return [])
+  | Error (`Msg msg) ->
+      Logs.err (fun m -> m "Failed to fetch jobs: %s" msg);
```
hannesm: If these three messages can be a bit more distinct, we'd have a better clue of what actually went wrong (`jobs_of_json`, `Utils.Json.from_string`, `Http.send_http_request`) -- so I propose to at least prefix these log messages with the function that went wrong.

builder_web.ml
```diff
@@ -313,83 +313,90 @@ let device_of_json = function
...
+      | Some (`Assoc latest) -> (
+          match Utils.Json.get "solo5_abi" latest with
+          | Some (`Assoc abi) -> (
+              match Utils.Json.get "target" abi with
+              | Some (`String "hvt") -> (
+                  match manifest_of_json (`Assoc latest) with
+                  | Ok manifest -> Ok (Some { name; synopsis; manifest })
+                  | Error e -> Error e)
+              | _ -> Ok None)
+          | _ -> Ok None)
```
reynir: I'm not sure you need to check the solo5 ABI is hvt. For qubes-miragevpn I get:

```shell
reynir@solsort:~$ solo5-elftool query-abi Downloads/qubes-miragevpn.xen 
{
  "type": "solo5.abi",
  "target": "xen",
  "version": 1
}
reynir@solsort:~$ solo5-elftool query-manifest Downloads/qubes-miragevpn.xen 
{
  "type": "solo5.manifest",
  "version": 1,
  "devices": [
  ]
}
```
So it will not fail. But of course it is not going to work all that well to try and deploy a xen unikernel. So the purpose is to skip things that are not hvt?

assets/main.js
PizieDust: in our form, our host devices are in a `select/option` thingy, and each option has a `value` (what is passed if you select the option` and a `label` what is displayed on the option in the UI. So I inspected it and the option and lavel are always the same value for this case so we don't need the `or`

builder_web.ml
PizieDust: Yes, we would like to only display `.hvt` vms in the form

unikernel.ml
PizieDust: Yes actually. Without it, parsing the multipart stream hangs trying to find "binary"

builder_web.ml
```diff
@@ -303,13 +303,14 @@ let compare_of_json = function
...
+type device = Network of string | Block of string
 type device_manifest = { uuid : string; devices : device list }
 
 let device_of_json = function
   | `Assoc xs -> (
       match (Utils.Json.get "name" xs, Utils.Json.get "type" xs) with
-      | Some (`String name), Some (`String type_) -> Ok { name; type_ }
+      | Some (`String name), Some (`String "NET_BASIC") -> Ok (Network name)
+      | Some (`String name), Some (`String "BLOCK_BASIC") -> Ok (Block name)
       | _ -> Error (`Msg "Invalid device JSON"))
```
reynir, line 314: Maybe add a case where both is ``Some `String _`` and report that we don't recognize the device type.

builder_web.ml
```diff
@@ -338,13 +339,16 @@ type job = { name : string; synopsis : string; manifest : device_manifest }
 
 let job_of_json = function
   | `Assoc xs -> (
-      let name =
-        match Utils.Json.get "name" xs with Some (`String s) -> s | _ -> ""
+      let* name =
+        match Utils.Json.get "name" xs with
+        | Some (`String s) -> Ok s
+        | _ -> Error (`Msg "No name in the json or name is not a string")
```
reynir:
```suggestion
        | _ -> Error (`Msg "No name in the job json or name is not a string")
```

builder_web.ml
```diff
@@ -338,13 +339,16 @@ type job = { name : string; synopsis : string; manifest : device_manifest }
...
       in
-      let synopsis =
+      let* synopsis =
         match Utils.Json.get "synopsis" xs with
-        | Some (`String s) -> s
-        | _ -> ""
+        | Some (`String s) -> Ok s
+        | _ ->
+            Error (`Msg "no synopsis in the json or synopsis is not a string")
```
reynir:
```suggestion
            Error (`Msg "no synopsis in the job json or synopsis is not a string")

File: builder_web.ml [hunk 25]
```diff
@@ -355,9 +359,22 @@ let job_of_json = function
...
+                      i
+                        "builder_web.jobs_of_json: no valid target type. \
+                         requires hvt unikernels.");
+                  Ok None)
+          | _ ->
+              Logs.info (fun i ->
+                  i
+                    "builder_web.jobs_of_json: no solo5_abi section in the \
+                     json.");
+              Ok None)
```
reynir: I'm not sure about these log messages. They will trigger for xen builds and for system packages, so this could be noisy. Maybe demote them to debug?

builder_web.ml
reynir:
```suggestion
              | Some (`String ("hvt" | "spt")) -> (

Comment (reynir) [reply builder_web.ml]: Hmm the suggestion looks weird to me. I think we can allow spt just as well.

File: builder_web.ml [hunk 26]
```diff
@@ -388,13 +405,15 @@ let fetch_unikernel_jobs http_client =
           match jobs_of_json json with
           | Ok jobs -> Lwt.return jobs
           | Error (`Msg e) ->
-              Logs.err (fun m -> m "Failed to fetch jobs: %s" e);
```
reynir: Would it make sense to use a builder-web log source here instead of the default? Then you can change the log level only for builder-web when you need to debug why a job isn't showing up. (This applies for the whole file)

unikernel.ml
```diff
@@ -1927,7 +1923,7 @@ struct
                                         ("Albatross JSON Error: " ^ err_msg)
                                         `Internal_server_error)
                             in
-                            if is_builder then
```
reynir: We could do a match instead:
```OCaml
match deploy_mode with
| "builder" -> ...
| "manual" -> ...
| _ -> generate_http_error_response "Bad deploy mode" `Bad_request
```

unikernel.ml
reynir: Ok. That makes sense. Maybe they should be two separate endpoints then? We can do that in another PR (if you agree please open an issue so we don't forget).

builder_web.ml
PizieDust: Can albatross deploy `spt` vms as well? In my understanding  I thought only `hvt` vms work with albatross?

builder_web.ml
PizieDust: This sounds great. Perhaps, we can set the logs to info, and then implement this in a new PR. Will create an issue referencing this.

builder_web.ml
```diff
@@ -355,25 +361,30 @@ let job_of_json = function
       | Some (`Assoc latest) -> (
           match Utils.Json.get "solo5_abi" latest with
           | Some (`Assoc abi) -> (
-              match Utils.Json.get "target" abi with
-              | Some (`String ("hvt" | "spt")) -> (
+              match Utils.Json.(get "target" abi, get "version" abi) with
+              | Some (`String abi_target), Some (`Int abi_version)
+                when String.equal abi_target "hvt"
+                     || String.equal abi_target "spt" -> (
```
reynir: We can avoid the `when` clause:
```suggestion
              | Some (`String ("hvt" | "spt" as abi_target)), Some (`Int abi_version) -> (
```
This helps the exhaustivity checker.

reynir (approved):

reynir (approved):

PR #234, opened 2026-05-06
Add timeout (1 hour) for background task, addresses #218

hannesm:
While a full solution would be to investigate where the task is stuck, cancelling it after one hour should be a fine first step.

### DIFF
--- a/unikernel.ml
+++ b/unikernel.ml
@@ -31,8 +31,8 @@ struct
 
   let recipient email =
     match Colombe_emile.to_path email with
-    | Ok path -> [ Colombe.Forward_path.Forward_path path ]
-    | Error (`Msg e) ->
+    | path -> [ Colombe.Forward_path.Forward_path path ]
+    | exception Invalid_argument e ->
         Logs.err (fun m ->
             m "Type conversion failed for %s: %s" (Emile.to_string email) e);
         []
@@ -71,8 +71,8 @@ struct
     in
     let sender =
       match Colombe_emile.to_path email_config.from_email with
-      | Ok path -> Some path
-      | Error _ -> None
+      | path -> Some path
+      | exception Invalid_argument _ -> None
     in
     let streamer =
       let s = Mrmime.Mt.to_stream email in
--- a/unikernel.ml
+++ b/unikernel.ml
@@ -2854,8 +2854,14 @@ struct
       Lwt.catch
         (fun () ->
           Logs.info (fun m -> m "Starting background update...");
-          run_background_update_check happy_eyeballs (Store.users store) stack
-            (Store.email store) !albatross_instances_ref http_client)
+          Lwt.choose
+            [
+              run_background_update_check happy_eyeballs (Store.users store)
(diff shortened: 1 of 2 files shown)


reynir (approved):

reynir:
I pushed a change. Colombe has a breaking change from 0.12.0 to 0.13.0 i colombe.emile. I upgraded instead of adding an upper bound in config.ml. Feel free to revert and add an upper bound instead. Good to merge now IMO.

hannesm:
Thanks @reynir. I wonder (asking @dinosaure) whether we need to catch the Invalid_argument exception, or can just remove the match? So, will `Colombe_emile.to_path` ever raise?

reynir:
When I looked at the diff in colombe it seems that the error case was converted into an exception. I didn't study further what the error cases might be. It's not documented in the .mli, and the commit only said something along "improve colombe.emile".

hannesm:
> error case was converted into an exception

Thanks! it would be good to get @dinosaure expertise on why this is now an exception... and indeed to me it looks very fine to handle the Invalid_argument... I'm just wondering whether this API can be made in such a way that we never get an exception.

unikernel.ml
```diff
@@ -31,8 +31,8 @@ struct
 
   let recipient email =
     match Colombe_emile.to_path email with
-    | Ok path -> [ Colombe.Forward_path.Forward_path path ]
-    | Error (`Msg e) ->
+    | path -> [ Colombe.Forward_path.Forward_path path ]
+    | exception Invalid_argument e ->
```
dinosaure, line 35: This case appears only for one specific case when the domain part of the email is a `Literal` (something like `foo@[foo]`). This case is **not handled** by `emile`, in other words, it's not possible to have an email address with a `Literal` domain if you use `Emile.of_string` (or equivalent). So you should never have such exception when you would like to transform an Emile address to a recipient.

I did this patch to facilitate the back and forth between Emile values and Colombe values.

hannesm: The path for email addresses is: the user supplies a string, and in Utils.Email the `t_of_json` uses `Mrmime.Mailbox.of_string` -- and then it is stored as a `Mrmime.Mailbox.t` and used to be passed to `Colombe_emile.to_path`

Now, my question is whether there could be an email that is provided that leads to an invalid argument.

On the other side, I guess we could directly store in `Utils.Email.t` a `Colombe.Path.t` (and then no need to bother here with potential exceptions, but only in the "read from disk", "register user", and "specify sender email address in configuration")... maybe for another PR.

dinosaure: > and in Utils.Email the t_of_json uses Mrmime.Mailbox.of_string 

Yes, `Mrmime.Mailbox.of_string` is an alias of `Emile.of_string`. So from what you explained, it's not possible to have an email address with a `Literal` domain. So when you use `Colombe_emile.to_path`, you will never get the `Invalid_argument` case :+1:.

And, for sure, I should improve the documentation about that.

PizieDust (approved):
thanks @hannesm

PR #235, opened 2026-05-06
unikernel_update: rephrase "Current Build" to "Deployed Unikernel" to make it obvious

hannesm:
The distinction is between what we have currently deployed and what is the latest build on our reproducible build infrastructure.

The previous terminology "Current Build" vs "Latest Build" is not obvious to me, but "Deployed Unikernel" is very obvious.


reynir (approved):

PizieDust (approved):
Thank you @hannesm .
I think this is fine to merge without the CI being green.

PR #236, opened 2026-05-07
cirrus: udpate to OCaml 4.14.3

hannesm:
I'm keeping this separate to figure out whether I can reproduce the issue that the CI for some reason insists on OCaml 5.4.1...


hannesm:
this is fine, so good to merge.

PR #239, opened 2026-05-10
revert part of #234, use older colombe -- to avoid opam-monorepo failures

hannesm:
I've no clue why opam-monorepo otherwise fails... :/


PizieDust (approved):
Thank you @hannesm

PR #241, opened 2026-05-21
Monitoring configuration in mollymawk

PizieDust:
In this PR we add a new network stack to mollymawk: called management.
To ensure names can resolve well, we have two new flags: 
- `management-dns-server`: which can be the ip of which dnsvizor is running on the management network as well, and
-  `management-host-name`: the hostname given to this dnsvizor

We then have two main api endpoints:
- `/api/unikernel/monitoring/status`: which checks if monitoring is available for the vm, and what log options and metrics are available
- `/api/unikernel/monitoring/update`: which allows mollymawk to send tcp commands to update the log options and metrics

We have a new file `monitoring.ml` which contains all the html and javascript needed to render the logs and metrics to the UI and also to call the two API endpoints above. We use `htmx` so we don't have to refresh the page.

The main functions contained here are:

* **`getColor(level)`**: Returns a background hex color code matching the specified log severity level (`info`, `warning`, `error`), falling back to dark gray.
* **`getTextColor(level)`**: Returns a high-contrast light text hex color code tailored to fit on top of the corresponding log level background.
* **`getCommand()` [Log version]**:  Generates the log configurations into a single, comma-separated command string (e.g., `L*:info,source:debug`).
* **`get allMetrics()`**: Evaluates to `true` if every single tracked metric in the dictionary is currently enabled, otherwise returns `false`.
* **`set allMetrics(value)`**: Simultaneously sets all individual metric toggles to either `true` or `false` when a master global switch is flipped.
* **`getCommand()` [Metric version]**: Generates the metrics configurations into a single comman-separated command string.

## Screenshots

image

image

image

image

### DIFF
--- a/config.ml
+++ b/config.ml
@@ -8,7 +8,8 @@ let management_stack, lease =
   generic_stackv4v6_with_lease ~group:"management" ~dhcp_requests
     (netif "management")
 
-let management_eyeballs = generic_happy_eyeballs ~group:"management" management_stack
+let management_eyeballs =
+  generic_happy_eyeballs ~group:"management" management_stack
 
 let management_dns =
   generic_dns_client ~group:"management" management_stack management_eyeballs
--- a/config.ml
+++ b/config.ml
@@ -8,7 +8,7 @@ let management_stack, lease =
   generic_stackv4v6_with_lease ~group:"management" ~dhcp_requests
     (netif "management")
 
-let management_eyeballs = generic_happy_eyeballs management_stack
+let management_eyeballs = generic_happy_eyeballs ~group:"management" management_stack
 
 let management_dns =
   generic_dns_client ~group:"management" management_stack management_eyeballs
--- a/config.ml
+++ b/config.ml
@@ -31,7 +31,8 @@ let mollymawk =
     ]
   in
   main ~packages "Unikernel.Main"
-    (stackv4v6 @-> stackv4v6 @-> kv_ro @-> block @-> http_client @-> job)
+    (stackv4v6 @-> stackv4v6 @-> dns_client @-> kv_ro @-> block @-> http_client
+   @-> job @-> job)
 
 let block = block_of_file "data"
 
@@ -45,13 +46,37 @@ let http_client =
     (tcpv4v6 @-> mimic @-> Mirage.http_client)
 
 let stack = generic_stackv4v6 default_network
-
-let management_stack =
-  generic_stackv4v6 ~group:"management" (netif "management")
-
 let eyeballs = generic_happy_eyeballs stack
 let dns = generic_dns_client stack eyeballs
 let tcp = tcpv4v6_of_stackv4v6 stack
+let dhcp_requests = make_dhcp_requests ()
+
+let management_stack, lease =
+  generic_stackv4v6_with_lease ~group:"management" ~dhcp_requests
+    (netif "management")
+
+let management_eyeballs = generic_happy_eyeballs management_stack
+
+let management_dns =
+  generic_dns_client ~group:"management" management_stack management_eyeballs
+
+let management_domain (dhcp_requests, lease) =
+  let () =
+    add_dhcp_request dhcp_requests 15
+    (* DOMAIN_NAME *)
+  in
+  let connect _ _ = function
+    | [ lease ] ->
+        (* TODO: sane default instead of None?! *)
+        code ~pos:__POS__
+          "Lwt.return @@@@@ match %s with None -> None@ | Some lease -> \
+           Dhcp_wire.find_domain_name lease"
+          lease
+    | _ -> assert false
+  in
+  impl ~connect ~extra_deps:[ dep lease ] "Dhcp_wire" job
+
+let management_domain_name = management_domain (dhcp_requests, lease)
 
 let http_client =
   let happy_eyeballs = mimic_happy_eyeballs stack eyeballs dns in
@@ -59,4 +84,7 @@ let http_client =
 
 let () =
   register "mollymawk"
-    [ mollymawk $ stack $ management_stack $ assets $ block $ http_client ]
+    [
+      mollymawk $ stack $ management_stack $ management_dns $ assets $ block
+      $ http_client $ management_domain_name;
+    ]
--- a/config.ml
+++ b/config.ml
@@ -2,53 +2,6 @@
 open Mirage
 
 let assets = crunch "assets"
-
-let mollymawk =
-  let packages =
-    [
-      package "logs";
-      package "x509";
-      package "tls-mirage";
-      package ~min:"2.7.0" "albatross";
-      package "yojson";
-      package "uri";
-      package "tyxml";
-      package "multipart_form";
-      package "mirage-crypto-rng";
-      package "uuidm";
-      package "emile";
-      package ~sublibs:[ "emile" ] ~min:"0.12.0" ~max:"0.13.0" "colombe";
-      package "sendmail";
-      package ~sublibs:[ "mirage" ] ~min:"0.5.0" "paf";
-      package "oneffs";
-      package "duration";
-      package ~min:"0.2.0" "ohex";
-      package "http-mirage-client";
-      package "multipart_form-lwt";
-      package "sendmail-mirage";
-      package "dns-client-mirage";
-      package "mrmime";
-    ]
-  in
-  main ~packages "Unikernel.Main"
-    (stackv4v6 @-> stackv4v6 @-> dns_client @-> kv_ro @-> block @-> http_client
-   @-> job @-> job)
-
-let block = block_of_file "data"
-
-let http_client =
-  let connect _ modname = function
-    | [ _tcpv4v6; ctx ] ->
-        code ~pos:__POS__ {ocaml|%s.connect %s|ocaml} modname ctx
-    | _ -> assert false
-  in
-  impl ~connect "Http_mirage_client.Make"
-    (tcpv4v6 @-> mimic @-> Mirage.http_client)
-
-let stack = generic_stackv4v6 default_network
-let eyeballs = generic_happy_eyeballs stack
-let dns = generic_dns_client stack eyeballs
-let tcp = tcpv4v6_of_stackv4v6 stack
 let dhcp_requests = make_dhcp_requests ()
 
 let management_stack, lease =
@@ -78,6 +31,54 @@ let management_domain (dhcp_requests, lease) =
 
 let management_domain_name = management_domain (dhcp_requests, lease)
 
+let mollymawk =
+  let packages =
+    [
+      package "logs";
+      package "x509";
+      package "tls-mirage";
+      package ~min:"2.7.0" "albatross";
+      package "yojson";
+      package "uri";
+      package "tyxml";
+      package "multipart_form";
+      package "mirage-crypto-rng";
+      package "uuidm";
+      package "emile";
+      package ~sublibs:[ "emile" ] ~min:"0.12.0" ~max:"0.13.0" "colombe";
+      package "sendmail";
+      package ~sublibs:[ "mirage" ] ~min:"0.5.0" "paf";
+      package "oneffs";
+      package "duration";
+      package ~min:"0.2.0" "ohex";
+      package "http-mirage-client";
+      package "multipart_form-lwt";
+      package "sendmail-mirage";
+      package "dns-client-mirage";
+      package "mrmime";
+    ]
+  in
+  main ~packages "Unikernel.Main"
+    ~deps:[ dep management_domain_name ]
+    (stackv4v6 @-> stackv4v6 @-> dns_client @-> kv_ro @-> block @-> http_client
+   @-> job)
+
+let block = block_of_file "data"
+
+let http_client =
+  let connect _ modname = function
+    | [ _tcpv4v6; ctx ] ->
+        code ~pos:__POS__ {ocaml|%s.connect %s|ocaml} modname ctx
+    | _ -> assert false
+  in
+  impl ~connect "Http_mirage_client.Make"
+    (tcpv4v6 @-> mimic @-> Mirage.http_client)
+
+let stack = generic_stackv4v6 default_network
+let eyeballs = generic_happy_eyeballs stack
+let dns = generic_dns_client stack eyeballs
+let tcp = tcpv4v6_of_stackv4v6 stack
+
 let http_client =
   let happy_eyeballs = mimic_happy_eyeballs stack eyeballs dns in
   http_client $ tcp $ happy_eyeballs
@@ -86,5 +87,5 @@ let () =
   register "mollymawk"
     [
       mollymawk $ stack $ management_stack $ management_dns $ assets $ block
-      $ http_client $ management_domain_name;
+      $ http_client;
     ]
--- a/config.ml
+++ b/config.ml
@@ -31,7 +31,7 @@ let mollymawk =
     ]
   in
   main ~packages "Unikernel.Main"
-    (stackv4v6 @-> kv_ro @-> block @-> http_client @-> job)
+    (stackv4v6 @-> stackv4v6 @-> kv_ro @-> block @-> http_client @-> job)
 
 let block = block_of_file "data"
 
@@ -45,6 +45,10 @@ let http_client =
     (tcpv4v6 @-> mimic @-> Mirage.http_client)
 
 let stack = generic_stackv4v6 default_network
+
+let management_stack =
+  generic_stackv4v6 ~group:"management" (netif "management")
+
 let eyeballs = generic_happy_eyeballs stack
 let dns = generic_dns_client stack eyeballs
 let tcp = tcpv4v6_of_stackv4v6 stack
@@ -54,4 +58,5 @@ let http_client =
   http_client $ tcp $ happy_eyeballs
 
 let () =
-  register "mollymawk" [ mollymawk $ stack $ assets $ block $ http_client ]
+  register "mollymawk"
+    [ mollymawk $ stack $ management_stack $ assets $ block $ http_client ]
--- a/monitoring.ml
+++ b/monitoring.ml
@@ -104,12 +104,8 @@ let render_metric_picker ~source ~value index =
     ]
 
 let monitoring_status_html ~name ~logs ~metrics =
-  let log_entries =
-    match logs with Error _ -> [] | Ok s -> parse_monitoring_response s
-  in
-  let metric_entries =
-    match metrics with Error _ -> [] | Ok s -> parse_monitoring_response s
-  in
+  let log_entries = parse_monitoring_response logs in
+  let metric_entries = parse_monitoring_response metrics in
   let default_log_level =
     match List.assoc_opt "*" log_entries with Some l -> l | None -> "info"
   in
--- a/monitoring.ml
+++ b/monitoring.ml
@@ -1,0 +1,409 @@
+open Tyxml_html
+
+let parse_monitoring_response str =
+  let str = String.trim str in
+  let str =
+    if String.length str > 4 && String.sub str 0 4 = "ok: " then
+      String.sub str 4 (String.length str - 4)
+    else str
+  in
+  String.split_on_char ',' str
+  |> List.filter_map (fun s ->
+      match String.split_on_char ':' (String.trim s) with
+      | [ name; value ] -> Some (String.trim name, String.trim value)
+      | _ -> None)
+
+let log_levels = [ "debug"; "info"; "warning"; "error"; "app"; "quiet" ]
+
+let render_log_picker ~source ~default_level ~value =
+  div
+    ~a:
+      [
+        a_class
+          [
+            "flex items-center justify-between gap-3 px-3 py-2 rounded \
+             bg-gray-700 hover:bg-gray-600 transition-colors";
+          ];
+      ]
+    [
+      span
+        ~a:
+          [
+            a_class [ "font-mono truncate flex-1 font-semibold" ];
+            a_title source;
+          ]
+        [
+          txt (if String.equal source "*" then "Default Log level" else source);
+        ];
+      select
+        ~a:
+          [
+            Unsafe.string_attrib "x-model" value;
+            Unsafe.string_attrib ":style"
+              (Fmt.str
+                 "`background-color: ${getColor(%s)}; color: \
+                  ${getTextColor(%s)}`"
+                 value value);
+            a_class
+              [
+                "text-xs border border-gray-600 rounded px-2 py-1 \
+                 focus:outline-none cursor-pointer shrink-0 rounded-xl";
+              ];
+          ]
+        (List.map
+           (fun l ->
+             option
+               ~a:
+                 ([ a_value l ]
+                 @ if l = default_level then [ a_selected () ] else [])
+               (txt l))
+           log_levels);
+    ]
+
+let render_metric_picker ~source ~value index =
+  label
+    ~a:
+      [
+        a_class
+          [
+            "flex items-center gap-2 cursor-pointer px-3 py-2 rounded \
+             transition-colors";
+          ];
+        Unsafe.string_attrib ":class"
+          (Fmt.str "%s ? 'bg-primary-900' : 'bg-gray-700 hover:bg-gray-600'"
+             value);
+      ]
+    [
+      input
+        ~a:
+          [
+            a_input_type `Checkbox;
+            a_id (Fmt.str "metric-%d" index);
+            a_name (Fmt.str "metric-%s-%d" source index);
+            a_class [ "accent-primary-500 shrink-0 cursor-pointer" ];
+            Unsafe.string_attrib "x-model" value;
+          ]
+        ();
+      span
+        ~a:[ a_class [ "font-semibold font-mono px-2 truncate" ] ]
+        [ txt source ];
+      span
+        ~a:
+          [
+            Unsafe.string_attrib "x-show" value;
+            a_class [ "ml-auto font-mono text-primary-400 shrink-0" ];
+          ]
+        [ txt "enabled" ];
+      span
+        ~a:
+          [
+            Unsafe.string_attrib "x-show" (Fmt.str "!%s" value);
+            a_class [ "ml-auto font-mono text-gray-500 shrink-0" ];
+          ]
+        [ txt "disabled" ];
+    ]
+
+let monitoring_status_html ~name ~logs ~metrics =
+  let log_entries =
+    match logs with Error _ -> [] | Ok s -> parse_monitoring_response s
+  in
+  let metric_entries =
+    match metrics with Error _ -> [] | Ok s -> parse_monitoring_response s
+  in
+  let default_log_level =
+    match List.assoc_opt "*" log_entries with Some l -> l | None -> "info"
+  in
+  let other_logs = List.filter (fun (s, _) -> s <> "*") log_entries in
+  let other_metrics = List.filter (fun (s, _) -> s <> "*") metric_entries in
+  let logs_dict =
+    String.concat ", "
+      (List.map
+         (fun (s, l) ->
+           Printf.sprintf "'%s': '%s'" (String.escaped s) (String.escaped l))
+         other_logs)
+  in
+  let metrics_dict =
+    String.concat ", "
+      (List.map
+         (fun (s, l) ->
+           let is_en = if String.equal l "enabled" then "true" else "false" in
+           Printf.sprintf "'%s': %s" (String.escaped s) is_en)
+         other_metrics)
+  in
+  let log_data =
+    Printf.sprintf
+      "{ advancedLogs: false, defaultLevel: '%s', logs: { %s }, \
+       getColor(level) { const colors = {'info': '#166534', 'warning': \
+       '#854d0e', 'error': '#7f1d1d'}; return colors[level] || '#27272a'; }, \
+       getTextColor(level) { const colors = {'info': '#bbf7d0', 'warning': \
+       '#fef08a', 'error': '#fecaca'}; return colors[level] || '#a1a1aa'; }, \
+       getCommand() { let cmd = 'L*:' + this.defaultLevel; for (const [k, v] \
+       of Object.entries(this.logs)) { cmd += ',' + k + ':' + v; }  return \
+       cmd; } }"
+      default_log_level logs_dict
+  in
+  let metric_data =
+    Printf.sprintf
+      "{ metrics: { %s }, get allMetrics() { return \
+       Object.values(this.metrics).every(v => v); }, set allMetrics(value) { \
+       for (const k in this.metrics) { this.metrics[k] = value; } }, \
+       getCommand() { const values = Object.values(this.metrics); if \
+       (values.every(v => v)) { return 'M*:enable'; } if (values.every(v => \
+       !v)) { return 'M*:disable'; } let cmd = 'M*:disable'; for (const [k, v] \
+       of Object.entries(this.metrics)) { if (v) { cmd += ',' + k + ':enable'; \
+       } } return cmd; } }"
+      metrics_dict
+  in
+  div
+    ~a:[ a_id "monitoring-container"; a_class [ "space-y-6 py-4" ] ]
+    [
+      (if log_entries = [] && metric_entries = [] then
+         div
+           ~a:[ a_class [ "text-center py-8" ] ]
+           [
+             i
+               ~a:
+                 [
+                   a_class
+                     [
+                       "fa-solid fa-circle-exclamation text-gray-500 text-2xl \
+                        mb-2";
+                     ];
+                 ]
+               [];
+             p
+               ~a:[ a_class [ "text-gray-400 mt-2" ] ]
+               [ txt "Monitoring unavailable for this unikernel." ];
+           ]
+       else div []);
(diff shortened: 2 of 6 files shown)


config.ml
```diff
@@ -31,7 +31,7 @@ let mollymawk =
     ]
   in
   main ~packages "Unikernel.Main"
-    (stackv4v6 @-> kv_ro @-> block @-> http_client @-> job)
+    (stackv4v6 @-> stackv4v6 @-> kv_ro @-> block @-> http_client @-> job)
```
reynir: You could add a `dns_client` here as well instead of constructing it in the unikernel. The benefit is you get the boot parameters "for free". Just remember to pass `~group:"management"` so the options don't collide with the other interface.

config.ml
```diff
@@ -45,6 +45,10 @@ let http_client =
     (tcpv4v6 @-> mimic @-> Mirage.http_client)
 
 let stack = generic_stackv4v6 default_network
+
+let management_stack =
+  generic_stackv4v6 ~group:"management" (netif "management")
+
 let eyeballs = generic_happy_eyeballs stack
 let dns = generic_dns_client stack eyeballs
```
reynir, line 80: Re: above comment.

You'd basically copy these two lines and use `management_stack` instead of `stack` and pass `~group:"management"` too.

monitoring.ml
```diff
@@ -0,0 +1,409 @@
+open Tyxml_html
+
+let parse_monitoring_response str =
+  let str = String.trim str in
+  let str =
+    if String.length str > 4 && String.sub str 0 4 = "ok: " then
+      String.sub str 4 (String.length str - 4)
+    else str
```
hannesm: what is the failure semantics? if there's no `ok: `, then the whole string is passed on here...

unikernel.ml
```diff
@@ -14,10 +14,25 @@ module K = struct
...
+        [ "management-dns-server" ]
+    in
+    Mirage_runtime.register_arg Arg.(required & opt (some string) None doc)
+
+  let management_host_name =
+    let doc =
+      Arg.info ~doc:"Host name for the management network"
+        [ "management-hostname" ]
+    in
+    Mirage_runtime.register_arg Arg.(required & opt (some string) None doc)
```
reynir: It's a bit unclear to me what this is for.

monitoring.ml
```diff
@@ -0,0 +1,409 @@
...
+  in
+  String.split_on_char ',' str
+  |> List.filter_map (fun s ->
+      match String.split_on_char ':' (String.trim s) with
+      | [ name; value ] -> Some (String.trim name, String.trim value)
+      | _ -> None)
```
hannesm: I wonder what an error response will look like, and what will happen here...

if the split on `:` doesn't result in a two-element list, that part is silently dropped...

unikernel.ml
```diff
@@ -14,10 +14,25 @@ module K = struct
   let port =
     let doc = Arg.info ~doc:"HTTP listen port." [ "port" ] in
     Mirage_runtime.register_arg Arg.(value & opt int 8080 doc)
+
+  let management_dns_server =
+    let doc =
+      Arg.info ~doc:"DNS server for the management network"
+        [ "management-dns-server" ]
+    in
+    Mirage_runtime.register_arg Arg.(required & opt (some string) None doc)
```
hannesm: I thought this information would flow via DHCP from dnsvizor?

unikernel.ml
hannesm: I'm not sure why you'd need this... and what is the value you're passing?

unikernel.ml
```diff
@@ -59,6 +76,28 @@ struct
             Ipaddr.Set.of_list [ Ipaddr.V6 addr ] |> Lwt.return_ok
         | Error e -> Lwt.return_error e)
 
+  let getaddrinfo_management dns : Management_HE.getaddrinfo =
+   fun record_type destination ->
+    match record_type with
+    | `A -> (
+        Management_Dns.gethostbyname dns destination >>= function
+        | Ok addr ->
+            Logs.info (fun m ->
```
reynir: I suspect we want to demote these to debug logging before merging, or? I imagine it gets noisy.

unikernel.ml
hannesm: Ah, it is more the domain name for the management network. We should as well get this parameter via DHCP (DHCP option Domain_name)..

It is used below in `Domain_name.of_string (unikernel_name ^ "." ^ K.management_host_name ())` -- and there we could use `Domain_name.prepend_label <dhcp-issued-domain-name> unikernel_name`... Now, I'm not 100% sure how we get the dhcp-issued-domain-name.

unikernel.ml
hannesm: Likely we've to move the dns_client to config.ml, and use the `generic_dns_client` there with the management_stack (see the dns example https://github.com/mirage/mirage-skeleton/tree/main/applications/dns)

unikernel.ml
PizieDust: When we deploy `dnsvizor` and pass a `--domain` flag. 
This value `management-hostname` should be the value of this flag so we can resolve IP addresses. 

if the domain is for instance `management` and we have a unikernel `unipi` then mollymawk is able to do `unipi.management` and get it's IP address.

PizieDust: When we deploy dnsvizor and pass a --domain flag.
This value management-hostname should be the value of this flag so we can resolve IP addresses.

if the domain is for instance management and we have a unikernel unipi then mollymawk is able to do unipi.management and get it's IP address.

hannesm: Maybe @reynir has an idea, it may need some tweaking of the mirage tool...

hannesm: > When we deploy `dnsvizor` and pass a `--domain` flag. This value `management-hostname` should be the value of this flag so we can resolve IP addresses.
> 
> if the domain is for instance `management` and we have a unikernel `unipi` then mollymawk is able to do `unipi.management` and get it's IP address.

Ok, thanks. Indeed this domain name `management` is transported by dnsvizor in the DHCP lease to mollymawk... so we should find a way to use that, instead of requiring the operator to pass the same string there as well.

unikernel.ml
```diff
@@ -1230,6 +1269,101 @@ struct
...
+      unikernel_name _token_or_cookie _user reqd =
+    let open Lwt.Infix in
+    let target_r =
+      match
+        Domain_name.of_string
+          (Configuration.name_to_str unikernel_name ^ "." ^ management_hostname)
+      with
+      | Ok dn -> Ok (Domain_name.to_string dn)
+      | Error e -> Error e
+    in
```
hannesm: That's a strange thing here -- so first you convert `unikernel_name` to a string, to append the `management_hostname`, then convert all of that to a `Domain_name.t` -- just to convert it back to a string?

..and then pass that string to happy eyeballs connect function...

unikernel.ml
```diff
@@ -1230,6 +1269,101 @@ struct
...
+          `Bad_request
+    | Ok target_str ->
+        let query cmd =
+          Management_HE.connect happy_eyeballs target_str [ 2323 ] >>= function
+          | Error (`Msg err) -> Lwt.return_error err
+          | Ok ((_ip, _port), flow) -> (
+              Management_S.TCP.write flow (Cstruct.of_string cmd) >>= function
+              | Error _ ->
+                  Management_S.TCP.close flow >>= fun () ->
+                  Lwt.return_error "Failed to write"
```
hannesm: in this error message and below, there is no context. is this displayed to the user? shown in the operator log?

would it be possible to add more context of which write failed, and what was attempted? i.e. `failed to query monitoring information from $<host|IP> with $<error-message>`?

unikernel.ml
```diff
@@ -1230,6 +1269,101 @@ struct
...
+    with
+    | Some (_, command), Some (_, unikernel_name) -> (
+        let open Lwt.Infix in
+        let target_r =
+          match
+            Domain_name.of_string (unikernel_name ^ "." ^ management_hostname)
+          with
+          | Ok dn -> Ok (Domain_name.to_string dn)
+          | Error e -> Error e
+        in
```
hannesm: same as above, in addition there's code duplication.

unikernel.ml
```diff
@@ -1230,6 +1269,101 @@ struct
...
+            | Ok ((_ip, _port), flow) -> (
+                Logs.info (fun m -> m "Passing the command %s" command);
+                Management_S.TCP.write flow (Cstruct.of_string command)
+                >>= function
+                | Error e ->
+                    Management_S.TCP.close flow >>= fun () ->
+                    Middleware.http_response ~api_meth:false reqd
+                      ~title:"Monitoring Error"
+                      ~data:(`String "Failed to write to monitoring port")
+                      `Internal_server_error
```
hannesm: missing some context: which "monitoring port" did you failed to write to?

unikernel.ml
```diff
@@ -1230,6 +1269,101 @@ struct
...
+              `Bad_request
+        | Ok target_str -> (
+            Management_HE.connect happy_eyeballs target_str [ 2323 ]
+            >>= function
+            | Error (`Msg err) ->
+                Middleware.http_response ~api_meth:false reqd
+                  ~title:"Monitoring Unavailable" ~data:(`String err)
+                  `Bad_request
+            | Ok ((_ip, _port), flow) -> (
+                Logs.info (fun m -> m "Passing the command %s" command);
```
hannesm: if you want to keep this, maybe on the debug level?

utils.ml
```diff
@@ -232,7 +232,7 @@ let csrf_form_input csrf =
         [
           a_input_type `Hidden;
           a_id "molly-csrf";
-          a_name "molly-csrf-input";
+          a_name "molly_csrf";
```
hannesm, line 235: not clear why this is part of this PR... maybe move it to a separate PR?

monitoring.ml
```diff
@@ -0,0 +1,409 @@
...
+  let log_data =
+    Printf.sprintf
+      "{ advancedLogs: false, defaultLevel: '%s', logs: { %s }, \
+       getColor(level) { const colors = {'info': '#166534', 'warning': \
+       '#854d0e', 'error': '#7f1d1d'}; return colors[level] || '#27272a'; }, \
+       getTextColor(level) { const colors = {'info': '#bbf7d0', 'warning': \
+       '#fef08a', 'error': '#fecaca'}; return colors[level] || '#a1a1aa'; }, \
+       getCommand() { let cmd = 'L*:' + this.defaultLevel; for (const [k, v] \
+       of Object.entries(this.logs)) { cmd += ',' + k + ':' + v; }  return \
+       cmd; } }"
```
hannesm: this looks like you construct the "command" (a string) in JavaScript, and then send it off from mollymawk?

if this is the case, that's slightly dangerous -- the client can send arbitrary data to the unikernel on the monitoring endpoint. I'm not sure whether (or when) this turns into a security issue - have to think about it.

unikernel.ml
reynir: It's a bit tedious, but in config.ml something along these lines:

```OCaml
(* add a job parameter to the [main] call *)
let dhcp_requests = make_dhcp_requests ()
let management_stack, lease =
  generic_stackv4v6_with_lease ~group:"management" ~dhcp_requests network
let management_domain (dhcp_requests, lease) =
  let () = add_dhcp_request dhcp_requests 15 (* DOMAIN_NAME *) in
  let connect _ _ = function
    | [ lease ] ->
      (* TODO: sane default instead of None?! *)
      code ~pos:__POS__
        "Lwt.return @@@@@ \
         match %s with None -> None@ \
        | Some lease -> Dhcp_wire.find_domain_name lease"
    | _ -> assert false
  in
  impl ~extra_deps:[ dep lease ]

(*  then apply management_domain (dhcp_requests, lease) before use *)
```

reynir: So using `job` is a bit ugly and we could introduce a "type" for this but in the end there's no guarantee with these "types" with the generated code...

reynir: Sorry, don't add a `job` parameter to the `Mirage.main` call - that requires a module and extending the unikernel functor with such a module parameter that we will just ignore anyway. Instead you can pass `~deps` with `management_domain (dhcp_requests, lease))`. Then in unikernel.ml write `let start ... management_domain = ...`

hannesm: thanks reynir, I guess the domain name via dhcp could very well be by default in the generic_dns_client, and then a resolve action could append the domain on demand (similar to how it works in /etc/resolv.conf on unix)... but that requires changes to dns and mirage, so let's first get it into shape here :)

unikernel.ml
PizieDust: I thought it was great to ensure the domain passed during boot works which is why I wanted to ensure it parses well. but since we now get the domain automatically from the dhcp lease i would be removing this block

config.ml
```diff
@@ -2,6 +2,34 @@
 open Mirage
 
 let assets = crunch "assets"
+let dhcp_requests = make_dhcp_requests ()
+
+let management_stack, lease =
+  generic_stackv4v6_with_lease ~group:"management" ~dhcp_requests
+    (netif "management")
+
+let management_eyeballs = generic_happy_eyeballs management_stack
```
reynir:
```suggestion
let management_eyeballs = generic_happy_eyeballs ~group:"management" management_stack
```

unikernel.ml
```diff
@@ -1230,6 +1255,78 @@ struct
     Middleware.http_response reqd ~title:"Unikernel Information"
       ~data:response_data `OK
 
+  let send_management_tcp_command happy_eyeballs ~command ~target =
+    let port = 2323 in
+    Management_HE.connect happy_eyeballs target [ port ] >>= function
+    | Error (`Msg err) ->
+        Logs.info (fun m -> m "Failed to connect to %s:%d" target port);
```
hannesm: this is a good improvement, but could there be the notion of "management" or "monitoring"? I'd actually name this function "send_monitoring_tcp_command"...

Also, an inconsistency across this function: here you Lwt.return_error err, but the `err` is not included in the log message.... in subsequent error cases you throw away the error message completely. IMHO it makes sense to always include it.

unikernel.ml
```diff
@@ -1230,6 +1255,78 @@ struct
...
+                Management_S.TCP.close flow >>= fun () ->
+                Lwt.return_ok (Cstruct.to_string cstruct)))
+
+  let unikernel_monitoring_status happy_eyeballs management_hostname
+      unikernel_name _token_or_cookie _user reqd =
+    let open Lwt.Infix in
+    let target =
+      Configuration.name_to_str unikernel_name ^ "." ^ management_hostname
+    in
+    send_management_tcp_command happy_eyeballs ~command:"l*" ~target
```
hannesm: I'd expect something to happen with an errorneous reply.

unikernel.ml
```diff
@@ -1230,6 +1255,78 @@ struct
...
+    >>= fun logs ->
+    send_management_tcp_command happy_eyeballs ~command:"m*" ~target
```
hannesm: here as well, error handling for the command failure?

monitoring.ml
```diff
@@ -0,0 +1,409 @@
...
+        [ txt "disabled" ];
+    ]
+
+let monitoring_status_html ~name ~logs ~metrics =
+  let log_entries =
+    match logs with Error _ -> [] | Ok s -> parse_monitoring_response s
+  in
+  let metric_entries =
+    match metrics with Error _ -> [] | Ok s -> parse_monitoring_response s
+  in
```
hannesm: that's not very nice error handling... you just throw away whatever error happened, which means all the work for constructing error messages was useless.

unikernel.ml
```diff
@@ -1230,6 +1255,78 @@ struct
...
+                Management_S.TCP.close flow >>= fun () ->
+                Logs.info (fun m ->
+                    m "Received eof while reading from %s:%d -> %s" target port
+                      command);
+                Lwt.return_error "EOF"
+            | Ok (`Data cstruct) ->
+                Management_S.TCP.close flow >>= fun () ->
+                Lwt.return_ok (Cstruct.to_string cstruct)))
+
+  let unikernel_monitoring_status happy_eyeballs management_hostname
```
hannesm: I'd name that argument "management_domain", not "management_hostname"...

unikernel.ml
```diff
@@ -1230,6 +1255,78 @@ struct
...
+    >>= fun metrics ->
+    let html =
+      Monitoring.monitoring_status_html
+        ~name:(Configuration.name_to_str unikernel_name)
+        ~logs ~metrics
+    in
+    let body = Format.asprintf "%a" (Tyxml_html.pp_elt ()) html in
+    reply reqd body `OK
+
+  let unikernel_monitoring_update happy_eyeballs management_hostname _user
```
hannesm: same, argument name...

unikernel.ml
```diff
@@ -1230,6 +1255,78 @@ struct
...
+      multipart_body reqd =
+    match
+      ( Map.find_opt "command" multipart_body,
+        Map.find_opt "unikernel_name" multipart_body )
+    with
+    | Some (_, command), Some (_, unikernel_name) -> (
```
hannesm, line 1337: here you still receive some string from a client and send it over TCP.... which allows a client to send anything there...

unikernel.ml
```diff
@@ -3261,15 +3372,27 @@ struct
...
+          Management_HE.create
+            ~getaddrinfo:(getaddrinfo_management management_dns)
+            management_stack
+        in
+        Logs.info (fun m ->
+            m "Domain from DHCP lease: %s"
+              (Option.value management_domain_name ~default:"none given"));
+        let management_hostname =
+          Option.value management_domain_name ~default:"management"
+        in
```
hannesm: these are not in sync, and I'd expect a decision:
- use `"management"` as domain if DHCP didn't provide any
- fail if DHCP didn't provide any

Also, here would be a good place to say "Domain_name.of_string_exn management_domain_name" -- since if DHCP-provided one is not a domain name, we can just exit early -- and if it is a proper one, we can pass a typed domain_name down (and don't need to use a string).

utils.ml
PizieDust: I am using htmx to do some of the API requests, which can update the UI without us needing to refresh (we can get a response and just swap some html with the response).

To do post request, we need to send our csrf token, previously, since we did everything in `main.js` we didn't care about the name and just used the id in javascript to get the value and in unikernel.ml we find a value `molly_crsf" which is passed in the json.

with htmx, it passes the name in the multipart form and the validation won't pass, as it expects a `molly_csrf` value but receives a `molly-csrf-input` value

PizieDust: I removed the commit, will move it to a new PR with some other changes

PizieDust: https://github.com/robur-coop/mollymawk/pull/242

monitoring.ml
```diff
@@ -0,0 +1,485 @@
...
+  else
+    let kind = String.sub command 0 1 in
+    let rest = String.sub command 1 (String.length command - 1) in
+    let segments = String.split_on_char ',' rest in
+    let is_valid_level s =
+      match String.trim s with
+      | "debug" | "info" | "warning" | "error" | "app" | "quiet" -> true
+      | _ -> false
+    in
+
```
hannesm: no need for this spurious newline

monitoring.ml
```diff
@@ -0,0 +1,485 @@
...
+        let value = String.sub segment 2 (String.length segment - 2) in
+        is_valid_val value
+      else
+        match String.split_on_char ':' segment with
+        | [ name; val_ ] -> is_valid_source name && is_valid_val val_
+        | _ -> false
+    in
+    match kind with
+    | "L" ->
+        if segments = [] || segments = [ "" ] then
```
hannesm: why do we need both cases? as far as I understand, `String.split_on_char` (which the the `segments` bindings) will never return an empty list -- even if the empty string is provided, a list with a single element (the empty string) will be returned.

monitoring.ml
```diff
@@ -0,0 +1,485 @@
...
+          Error "Empty logs command body"
+        else if List.for_all (validate_segment is_valid_level) segments then
+          Ok command
+        else Error "Invalid logs configuration formatting"
+    | "M" ->
+        if segments = [] || segments = [ "" ] then
```
hannesm: see above

monitoring.ml
```diff
@@ -0,0 +1,485 @@
...
+              loop (i + 1)
+          | _ -> false
+      in
+      loop 0
+    in
+    let validate_segment is_valid_val segment =
+      let segment = String.trim segment in
+      if String.length segment = 0 then false
+      else if segment.[0] = '*' then
+        let value = String.sub segment 2 (String.length segment - 2) in
```
hannesm: here some length check is required, otherwise on the input `*`, it will raise an exception

hannesm: also on `*:`, or `**`

monitoring.ml
```diff
@@ -0,0 +1,485 @@
...
+    let is_valid_level s =
+      match String.trim s with
+      | "debug" | "info" | "warning" | "error" | "app" | "quiet" -> true
+      | _ -> false
+    in
+
+    let is_valid_metric s =
+      match String.trim s with "enable" | "disable" -> true | _ -> false
+    in
+    let is_valid_source name =
```
hannesm: I'm not entirely sure what restrictions are imposed for a source, esp. with the current state of logs and metrics, it seems that a source name may be any arbitrary string... I suggest to add a TODO somewhere that as source name everything is allowed. related https://github.com/dbuenzli/logs/issues/63

monitoring.ml
```diff
@@ -0,0 +1,485 @@
...
+  let command = String.trim command in
+  if String.length command < 2 then Error "Command is too short"
+  else
+    let kind = String.sub command 0 1 in
+    let rest = String.sub command 1 (String.length command - 1) in
+    let segments = String.split_on_char ',' rest in
+    let is_valid_level s =
+      match String.trim s with
+      | "debug" | "info" | "warning" | "error" | "app" | "quiet" -> true
+      | _ -> false
```
hannesm: you could as well use `Result.is_ok (Logs.level_of_string s)`

monitoring.ml
```diff
@@ -0,0 +1,485 @@
...
+        | [ name; val_ ] -> is_valid_source name && is_valid_val val_
+        | _ -> false
+    in
+    match kind with
+    | "L" ->
+        if segments = [] || segments = [ "" ] then
+          Error "Empty logs command body"
+        else if List.for_all (validate_segment is_valid_level) segments then
+          Ok command
+        else Error "Invalid logs configuration formatting"
```
hannesm: it may help for debugging purposes at least to expose the invalid string -- though it may need to be escaped (if rendered as html, etc.)

monitoring.ml
```diff
@@ -0,0 +1,485 @@
...
+    | "M" ->
+        if segments = [] || segments = [ "" ] then
+          Error "Empty metrics command body"
+        else if List.for_all (validate_segment is_valid_metric) segments then
+          Ok command
+        else Error "Invalid metrics configuration formatting"
```
hannesm: see above

monitoring.ml
```diff
@@ -0,0 +1,485 @@
...
+    | _ -> Error "Invalid command prefix (must be L or M)"
```
hannesm: in this function, there is a lot of `String.trim` -- why is this needed? I'm already confused why an initial `let command = String.trim command` is needed, but then each source name and value (log level/metrics enabled or disabled) is first `String.trim`...

monitoring.ml
```diff
@@ -0,0 +1,485 @@
...
+
+let check_monitoring_response_format str =
+  let str = String.trim str in
```
hannesm, line 68: again, I'm surprised by the String.trim.

monitoring.ml
```diff
@@ -0,0 +1,485 @@
...
+  if String.length str > 4 && String.sub str 0 4 = "ok: " then
+    Ok (String.sub str 4 (String.length str - 4))
+  else
+    Error
+      (Fmt.str
+         "Response from monitoring does not start with \"ok: \". Received: %s"
```
hannesm, line 74: Maybe use `%S` here to escape it (and avoid potential cross-site scripting)?

monitoring.ml
```diff
@@ -0,0 +1,485 @@
...
+let split_key_value_sources str =
+  String.split_on_char ',' str
+  |> List.fold_left
+       (fun acc s ->
+         match acc with
+         | Error err -> Error err
+         | Ok lst -> (
+             match String.split_on_char ':' (String.trim s) with
+             | [ name; value ] ->
+                 Ok ((String.trim name, String.trim value) :: lst)
```
hannesm: again, why are all the String.trim needed? what is the data that you encounter that has all these whitespace in there?

monitoring.ml
```diff
@@ -0,0 +1,485 @@
...
+             | _ -> Error "Invalid key-value source format"))
```
hannesm: Here, I'd again expose the data `s` (as `String.escaped s`).

monitoring.ml
```diff
@@ -0,0 +1,485 @@
...
+
+let parse_monitoring_response str =
+  match check_monitoring_response_format str with
+  | Ok str -> (
+      match split_key_value_sources str with
+      | Ok lst -> Ok lst
+      | Error err -> Error err)
+  | Error err -> Error err
+
+let log_levels = [ "debug"; "info"; "warning"; "error"; "app"; "quiet" ]
```
hannesm: this list of valid log levels is used in `check_command` -- maybe bind it earlier to use it there as well?

monitoring.ml
PizieDust: my understanding is String.trim only removes trailing whitespace..so.
like, if we have a string "This is a  word "
it will remove the white space after `word` but not the one after the `a`..

monitoring.ml
PizieDust: normally, we have no whitespace, but since it's constructed with Javascript and sent, if someone were to tamper with the payload...

unikernel.ml
```diff
@@ -1230,6 +1255,129 @@ struct
...
+    let port = 2323 in
+    let target = target ^ "." ^ Domain_name.to_string management_domain in
+    Management_HE.connect happy_eyeballs target [ port ] >>= function
+    | Error (`Msg err) ->
+        Logs.info (fun m ->
+            m "Failed to connect to %s on port %d with error: %s" target port
+              err);
+        Lwt.return_error
+          (Fmt.str "Failed to connect to %s on port %d with error: %s" target
+             port err)
```
hannesm: what about:
```OCaml
let msg = Fmt.str "[monitoring] Failed to connect to %s:%d, error %s" target port err in
Logs.info (fun m -> m msg);
Lwt.return_error msg
```

unikernel.ml
```diff
@@ -1230,6 +1255,129 @@ struct
...
+          Monitoring.monitoring_status_html
+            ~name:(Configuration.name_to_str unikernel_name)
+            ~logs:logs_data ~metrics:metrics_data
+        in
+        let body = Format.asprintf "%a" (Tyxml_html.pp_elt ()) html in
+        reply reqd body `OK
+    | Error err, _ ->
+        reply reqd
+          (Fmt.str "<p class=\"text-secondary-500\">%s</p>" err)
+          `Bad_request
```
hannesm: in other parts, `Middleware.http_response` is used -- here it is `reply` -- why?

PizieDust: with htmx, the response we receive is swapped in as html where we want, and with `reply` we can send html. `Middleware.http_response` expects json..

monitoring.ml
hannesm: if someone wants to tamper with the payload, they better make sure it is still well-formed... I'm not a big fan of allowing whitespaces at every place here...

so, from my point: better ensure our own JavaScript doesn't emit arbitrary whitespaces, and then disallow `foo : debug   ` -- and be very strict what to accept. :)

unikernel.ml
hannesm: maybe worth to include a comment?

monitoring.ml
```diff
@@ -1,17 +1,25 @@
...
+  [
+    Some Logs.Debug;
+    Some Logs.Info;
+    Some Logs.Warning;
+    Some Logs.Error;
+    Some Logs.App;
+  ]
+
+let log_level_of_string s =
+  match Logs.level_of_string s with Ok v -> Ok v | Error err -> Error err
```
hannesm: Why `match`? In both cases it returns the value it got. This could be `let log_level_of_string s = Logs.level_of_string s` -- but that's a bit pointless, so it can be completely removed and the callers can use `Logs.level_of_string` directly.

monitoring.ml
```diff
@@ -1,17 +1,25 @@
 open Tyxml_html
 
+let log_levels =
+  [
+    Some Logs.Debug;
+    Some Logs.Info;
+    Some Logs.Warning;
+    Some Logs.Error;
+    Some Logs.App;
```
hannesm, line 9: why "Some" in all cases? I'd use `Logs.[ Debug ; Info ; Warning ; Error ; App ]`.

monitoring.ml
```diff
@@ -180,17 +186,31 @@ let render_metric_picker ~source ~value index =
     ]
 
 let render_log_sources log_entries unikernel_name =
+  let parsed_log_entries =
+    List.map
+      (fun (s, l) ->
+        let parsed_level =
+          match Logs.level_of_string l with
+          | Ok level -> level
+          | Error _ -> Some Logs.Info
```
hannesm: huh? Why drop the error and replace it with some static value?

monitoring.ml
```diff
@@ -14,7 +14,8 @@ let log_level_of_string s =
 
 let check_command command =
   let command = String.trim command in
-  if String.length command < 2 then Error "Command is too short"
+  if String.length command < 2 then
+    Error (Fmt.str "Command is too short. got %s" command)
```
hannesm: Either `%S` or `(String.escaped command)`.

hannesm: well, actually, I'm not sure where this message is used, eventually there's more needed such as escaping `<` and `>` and `&`

monitoring.ml
```diff
@@ -50,18 +51,20 @@ let check_command command =
     in
     match kind with
     | "L" ->
-        if segments = [] || segments = [ "" ] then
-          Error "Empty logs command body"
+        if segments = [ "" ] then
+          Error (Fmt.str "Empty logs command body. got %s" command)
```
hannesm: see above

monitoring.ml
```diff
@@ -50,18 +51,20 @@ let check_command command =
...
         else if List.for_all (validate_segment is_valid_level) segments then
           Ok command
-        else Error "Invalid logs configuration formatting"
+        else Error (Fmt.str "Invalid logs configuration. got %s" command)
```
hannesm: see above

monitoring.ml
```diff
@@ -50,18 +51,20 @@ let check_command command =
...
     | "M" ->
-        if segments = [] || segments = [ "" ] then
-          Error "Empty metrics command body"
+        if segments = [ "" ] then
+          Error (Fmt.str "Empty metrics command body. got %s" command)
```
hannesm: see above

monitoring.ml
```diff
@@ -50,18 +51,20 @@ let check_command command =
...
         else if List.for_all (validate_segment is_valid_metric) segments then
           Ok command
-        else Error "Invalid metrics configuration formatting"
-    | _ -> Error "Invalid command prefix (must be L or M)"
+        else Error (Fmt.str "Invalid metrics configuration. got %s" command)
```
hannesm: see above

monitoring.ml
```diff
@@ -50,18 +51,20 @@ let check_command command =
...
+    | _ ->
+        Error
+          (Fmt.str "Invalid command prefix (must be L or M). got %s" command)
```
hannesm: see above

monitoring.ml
```diff
@@ -70,7 +73,7 @@ let check_monitoring_response_format str =
   else
     Error
       (Fmt.str
-         "Response from monitoring does not start with \"ok: \". Received: %s"
+         "Response from monitoring does not start with \"ok: \". Received: %S"
```
hannesm: see above

monitoring.ml
```diff
@@ -80,10 +83,12 @@ let split_key_value_sources str =
...
-             match String.split_on_char ':' (String.trim s) with
-             | [ name; value ] ->
-                 Ok ((String.trim name, String.trim value) :: lst)
-             | _ -> Error "Invalid key-value source format"))
+             match String.split_on_char ':' s with
+             | [ name; value ] -> Ok ((name, value) :: lst)
+             | _ ->
+                 Error
+                   (Fmt.str "Invalid key-value source format. got %s"
+                      (String.escaped s))))
```
hannesm: see above

unikernel.ml
```diff
@@ -1230,6 +1256,123 @@ struct
     Middleware.http_response reqd ~title:"Unikernel Information"
       ~data:response_data `OK
 
+  let send_monitoring_tcp_command happy_eyeballs ~command ~target
```
hannesm, line 1259: error strings produced by this function are embedded below in a `<p class..>%s</p>` -- so I guess to avoid XSS there's some escaping needed (or use of TyXML).

But then, the `commands` used  are only `l*` and `m*` here -- and the TCP errors shouldn't include any HTML entities, so it is likely fine.

hannesm:
looks fine to me now, thanks

PR #242, opened 2026-05-23
Rename name value from molly-csrf-input  to molly_csrf

PizieDust:
This change is needed so when we use htmx to send post request the right name for checking the csrf token on the backend is used.


hannesm:
looks good to me, thanks!

PR #243, opened 2026-05-23
Make modal content scrollable and to fit better

PizieDust:
This PR is to make modals more responsive to the inner content and to enable scrolling within the modal


hannesm:
looks fine to me, although I don't have much knowledge of this html stuff... feel free to merge when you're happy with it :)

PR #244, opened 2026-05-25
address review comments

hannesm:
Dear @PizieDust, here are some commits (PRed against the monitoring PR) to address review comments I put in earlier today. Please have a look and when you're fine with them, merge them (which well then show up in #241).

### DIFF
--- a/monitoring.ml
+++ b/monitoring.ml
@@ -9,10 +9,18 @@ let log_levels =
     Some Logs.App;
   ]
 
+let escape_string s =
+  String.escaped
+    (String.concat "&amp;"
+       (String.split_on_char '&'
+          (String.concat "&gt;"
+             (String.split_on_char '>'
+                (String.concat "&lt;" (String.split_on_char '<' s))))))
+
 let check_command command =
   let command = String.trim command in
   if String.length command < 2 then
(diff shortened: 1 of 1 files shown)


monitoring.ml
```diff
@@ -9,13 +9,18 @@ let log_levels =
     Some Logs.App;
   ]
 
-let log_level_of_string s =
-  match Logs.level_of_string s with Ok v -> Ok v | Error err -> Error err
+let escape_string s =
```
hannesm: I actually think we don't need this... all the errors seem to be sent back as HTML only inside of well-typed TyXML stuff, which should properly escape HTML entities. All that is needed here is `let escape_string = String.escaped`.

monitoring.ml
```diff
@@ -9,13 +9,18 @@ let log_levels =
...
+  String.escaped
+    (String.concat "&amp;"
+       (String.split_on_char '&'
+          (String.concat "&gt;"
+             (String.split_on_char '>'
+                (String.concat "&lt;" (String.split_on_char '<' s))))))
```
hannesm:
```suggestion
let escape_string = String.escaped
```

PizieDust (approved):
Thank you @hannesm

PR #245, opened 2026-05-27
Add scaling policies

PizieDust:
This PR adds scaling policies, which will be used to determine:
- should we scale this unikernel,
- max instances (default: 1)
- min instances (default: 1)

### DIFF
--- a/user_model.ml
+++ b/user_model.ml
@@ -67,7 +67,47 @@ let unikernel_update_to_json (u : unikernel_update) : Yojson.Basic.t =
       ("timestamp", `String (Utils.TimeHelper.string_of_ptime u.timestamp));
     ]
 
+let scaling_policy_to_json (p : unikernel_scaling_policy) : Yojson.Basic.t =
+  `Assoc
+    [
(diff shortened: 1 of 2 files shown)


user_model.ml
```diff
@@ -29,6 +29,13 @@ type unikernel_update = {
   timestamp : Ptime.t;
 }
 
+type unikernel_scaling_policy = {
+  name : Vmm_core.Name.Label.t;
+  primary_albatross_instance : Vmm_core.Name.Label.t;
+  max_instances : int;
+  should_scale : bool;
+}
```
reynir, line 37: Hmm, I suspect we more likely want a map from name * instance to scaling policy. Otherwise, if in the list we have two values with the same name and instance what do we do? We can likely enforce this invariant otherwise though.

reynir (approved):
Looks fine, but the scaling policies are not being used yet.

user_model.ml
PizieDust: in #246 we ensure that during crud operations we don't have duplicates

PR #246, opened 2026-05-27
Scaling policies UI

PizieDust:
This PR builds on #245 with the UI and api functionality to update scaling policies.

### DIFF
--- a/unikernel.ml
+++ b/unikernel.ml
@@ -1416,45 +1416,49 @@ struct
     in
     match Map.find_opt "max_instances" multipart_body with
     | Some (_, max_instances_str) -> (
-        let max_instances =
-          match int_of_string_opt max_instances_str with
-          | Some n -> n
-          | None -> 1
-        in
-        if max_instances = 1 then
-          let scaling_policies = remove_scaling_policy () in
-          update_unikernel_scaling scaling_policies
-        else
-          user_max_allowed_unikernel_instances stack albatross user.name
-          >>= function
-          | Error err ->
-              reply reqd
-                (Utils.display_alert
-                   (Fmt.str "An error occurred: %s" err)
-                   `Error)
-                `Bad_request
-          | Ok max_allowed ->
-              if max_instances > max_allowed then
-                reply reqd
-                  (Utils.display_alert
-                     (Fmt.str "Error: you can spawn a maximum of %d clones."
-                        max_allowed)
-                     `Error)
-                  `Bad_request
-              else
-                let new_policy =
-                  match current_scaling_policy with
-                  | Some p -> { p with max_instances }
-                  | None ->
-                      {
-                        User_model.name = unikernel_name;
-                        primary_albatross_instance =
-                          albatross.Albatross.configuration.name;
-                        max_instances;
-                      }
-                in
-                let scaling_policies = new_policy :: remove_scaling_policy () in
-                update_unikernel_scaling scaling_policies)
+        match int_of_string_opt max_instances_str with
+        | Some max_instances when max_instances >= 1 -> (
+            if max_instances = 1 then
+              let scaling_policies = remove_scaling_policy () in
+              update_unikernel_scaling scaling_policies
+            else
+              user_max_allowed_unikernel_instances stack albatross user.name
+              >>= function
+              | Error err ->
+                  reply reqd
+                    (Utils.display_alert
+                       (Fmt.str "An error occurred: %s" err)
+                       `Error)
+                    `Bad_request
+              | Ok max_allowed ->
+                  if max_instances > max_allowed then
+                    reply reqd
+                      (Utils.display_alert
+                         (Fmt.str "Error: you can spawn a maximum of %d clones."
+                            max_allowed)
+                         `Error)
+                      `Bad_request
+                  else
+                    let new_policy =
+                      match current_scaling_policy with
+                      | Some p -> { p with max_instances }
+                      | None ->
+                          {
+                            User_model.name = unikernel_name;
+                            primary_albatross_instance =
+                              albatross.Albatross.configuration.name;
+                            max_instances;
+                          }
+                    in
+                    let scaling_policies =
+                      new_policy :: remove_scaling_policy ()
+                    in
+                    update_unikernel_scaling scaling_policies)
+        | None | Some _ ->
+            reply reqd
+              (Utils.display_alert "Max instances must be greater than 1."
+                 `Error)
+              `Bad_request)
     | _ ->
         reply reqd
           (Utils.display_alert "Missing max number of clones to spawn." `Error)
--- a/unikernel.ml
+++ b/unikernel.ml
@@ -544,6 +544,22 @@ struct
             (instance.configuration.name, []))
       (Albatross.Albatross_map.bindings albatross_instances)
 
+  let user_max_allowed_unikernel_instances stack albatross user_name =
+    user_unikernels_by_instance stack albatross user_name >>= fun unikernels ->
+    let res =
+      match Albatross_state.policy albatross ~domain:user_name with
+      | Error err -> Error err
+      | Ok user_albatross_policy ->
+          let max_allowed =
+            Option.fold ~none:0
+              ~some:(fun (user_policy : Vmm_core.Policy.t) ->
+                user_policy.unikernels - List.length unikernels)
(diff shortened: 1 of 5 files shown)


unikernel.ml
```diff
@@ -544,6 +544,22 @@ struct
...
+    user_unikernels_by_instance stack albatross user_name >>= fun unikernels ->
+    let res =
+      match Albatross_state.policy albatross ~domain:user_name with
+      | Error err -> Error err
+      | Ok user_albatross_policy ->
+          let max_allowed =
+            Option.fold ~none:0
+              ~some:(fun (user_policy : Vmm_core.Policy.t) ->
+                user_policy.unikernels - List.length unikernels)
+              user_albatross_policy
```
reynir, line 556: Is this correct? If there's no policy we are not allowed to create any unikernels?

unikernel.ml
```diff
@@ -1364,6 +1380,97 @@ struct
...
+    in
+    match
+      ( Map.find_opt "should_scale" multipart_body,
+        Map.find_opt "max_instances" multipart_body )
+    with
+    | Some _, Some (_, max_instances_str) -> (
+        let max_instances =
+          match int_of_string_opt max_instances_str with
+          | Some n -> n
+          | None -> 1
```
reynir: Shouldn't we rather error out if we get something that is not a number?

unikernel.ml
```diff
@@ -1364,6 +1380,97 @@ struct
...
+        in
+        if max_instances = 1 then
```
reynir: what if `max_instances` is 0 or negative

unikernel.ml
```diff
@@ -1364,6 +1380,97 @@ struct
...
+                `Bad_request
+          | Ok max_allowed ->
+              if max_instances > max_allowed then
+                reply reqd
+                  (Utils.display_alert
+                     (Fmt.str "Error: you can spawn a maximum of %d clones."
+                        max_allowed)
+                     `Error)
+                  `Bad_request
+              else
```
reynir: Not sure about this logic. So `max_allowed` is the maximum allowed **minus** the number of running unikernels. So the max_instances we can at most set to that number. But what if we remove some of the running unikernels? What if some of the running unikernels are auto scaled unikernels that are then automatically removed? I think it makes more sense to check against what the policy says.

Regardless, we will have the issue that perhaps multiple unikernels need to auto scale up and we might not in the policy have enough unikernels for that.

unikernel.ml
```diff
@@ -1364,6 +1380,97 @@ struct
...
+                let new_policy =
+                  match current_scaling_policy with
+                  | Some p -> { p with max_instances; should_scale = true }
```
reynir: Is `should_scale` ever not `true`?

unikernel.ml
```diff
@@ -1364,6 +1380,102 @@ struct
...
           `Bad_request
 
+  let unikernel_scaling_policy_update stack store albatross unikernel_name
+      (user : User_model.user) multipart_body reqd =
+    let current_scaling_policy =
+      List.find_opt
+        (fun (scaling_policy : User_model.unikernel_scaling_policy) ->
+          Vmm_core.Name.Label.equal unikernel_name scaling_policy.name
+          && Vmm_core.Name.Label.equal albatross.Albatross.configuration.name
+               scaling_policy.primary_albatross_instance)
```
reynir, line 1389: I think it's unclear to me what `primary_albatross_instance` is meant to be used for.

unikernel.ml
PizieDust: yes this is correct.

unikernel.ml
PizieDust: I will add a check for this and error out

unikernel.ml
PizieDust: so we "may" have different albatross instances, for example:
albatross-1 and albatross-2...
which means we may have a unikernel deployed on albatross-1 as hello, and another deployed on albatross-2 as hello. 

so to prevent being able to delete 1 or the other, or add a policy to the wrong one, we have to add the instance name to the list, that way we filter by the unikernel name and by the albatross instance name.

now, it's possible in the future that during scaling, we may scale hello from albatross-1 by creating a clone in albatross-2, which is why we need primary-albatross-instance so we know where the original vm was deployed in case the scaling policy changes, 

the idea is that the original vm and clones share the same scaling policy, and eventually, when we tear down clones, we should know which is the original

unikernel.ml
PizieDust: originally i just wanted a should_scale or not, but now i think the design evovled to be that only vms with a scaling policy are in the user's list, so in a way should_scale is now useless. so i'll remove it's usage.

unikernel.ml
PizieDust: > Not sure about this logic. So `max_allowed` is the maximum allowed **minus** the number of running unikernels. 

This is correct, we want scaling to be bounded by the users allowed policy

> So the max_instances we can at most set to that number. But what if we remove some of the running unikernels? 

if some of the running unikernels are removed, then the user's max_allowed increases proportionally, and they can increase the max_instances of any other

> What if some of the running unikernels are auto scaled unikernels that are then automatically removed? I think it makes more sense to check against what the policy says.

I think we already do this in the function `user_max_allowed_unikernel_instances`. We get the users policy minus how many unikernels they already have running to get their max_instances. 

> Regardless, we will have the issue that perhaps multiple unikernels need to auto scale up and we might not in the policy have enough unikernels for that.

This is true. In this scase, I think we in the future, send the user an email about which unikernels were able to scale and which ones needed to scale but didn't due to insufficient resources

unikernel.ml
reynir: With the change addressing this you removed the possibility to remove a policy https://github.com/robur-coop/mollymawk/commit/b02aac479842378fa4ac3537d10b5a7e00398449#commitcomment-186774303

But maybe removing policies should be a different endpoint.

unikernel.ml
reynir: I'm not convinced we need to subtract how many unikernels we have running already. That might change at any point, both up and down.
- Say you have a policy allowing 16 unikernels. At the moment of setting scaling policy you have 8 unikernels running. You then set the max allowed to 8 (the maximum allowed). So the logic is still missing, but if you count the original against the max allowed you can only auto spin up 7 more unikernels leaving one unikernel slot unused.
- Four of the 8 already running unikernels are temporary jobs, and they shut down. Now you have 12 available unikernel slots, but the auto scaling logic is only able to use the 8 slots because at the time when the autoscaling policy was set that was what was unused.
- Some time later auto scaling triggers and we now have 11 running unikernels. The temporary jobs are then started and we have 15 running unikernels. We want to set an autoscaling policy for a different unikernel B, but we only have 1 available slot so we can only set max_allowed to 1 even though 4 of the unikernels are temporary and another 7 are auto scaled unikernels that might get collected at any point.

So with that in mind I think we should allow up to what is the maximum allowed by the policy and ignore how many unikernels are actually running right now (because that number can easily change).

unikernel.ml
PizieDust: I just added back passing should_scale in the API request and using that to determine if to remove a policy, but not save the should_scale value. 
since should_scale on the UI is a toggle, we can say if the toggle is off, then the user has intended to remove the policy. this way we handle both type of actions with one endpoint

unikernel.ml
PizieDust: I added this in the latest commit.
I think albatross policies may affect scaling in general. 

say the policy gives a user 10 unikernels, and they have 4 deployed unikernels.

now the scaling policy, we say they can set the max_allowed to the original policy which is 10. 

now if a scaling is to happen, it can only scale 6 vms in reality if the other 4 are still running, as albatross won't allow them have more than 10 running unikernels.

but I think there's no easy way to handle this cleanly

reynir: > now if a scaling is to happen, it can only scale 6 vms in reality if the other 4 are still running, as albatross won't allow them have more than 10 running unikernels.
> 
> but I think there's no easy way to handle this cleanly

Yes, it's a situation we have to deal with in any case.

reynir (approved):
Looks good to me now. I have not looked at the html.

PizieDust:
> Looks good to me now. I have not looked at the html.

@reynir  It looks like this:

image

PR #250, opened 2026-05-29
Add autoscaling module

PizieDust:
This module implements the logic for monitoring CPU usage of the currently running unikernels and decides when to scale up or scale down.

## Configuration

* **`poll_interval`**: How frequently the system expects to evaluate the unikernel stats.
* **`scale_up_threshold_percent`**: The CPU load threshold that triggers a scale up.
* **`scale_up_trigger_ticks`**: How many consecutive stats reports must exceed the scale up threshold before a clone is created.
* **`scale_down_threshold_percent`**: The CPU load threshold that triggers a scale down event.
* **`scale_down_trigger_ticks`**: How many consecutive stats reports must fall below the scale down threshold before a clone is destroyed.
* **`cooldown_period`**: The grace period after any scaling action where the system pauses scaling to prevent rapid creation/destruction of clones.
* **`death_timeout`**: The maximum time to wait for stats from a VM before it is considered dead.

### `Cpu_monitor`

Converts raw `rusage` stats from Albatross into a CPU utilization percentage (0-100%).

### `Cluster_manager`

Tracks and evaluates load metrics for groups containing a primary unikernel and all of its clones.

* **`in_cooldown`**: Checks if the group recently had a scale operation and should ignore load changes.
* **`extract_name_and_clone_id`**: Parses a clone's name to separate the primary name from the clone ID.
* **`create_group`**: Initializes a new scaling group state with the primary VM.
* **`next_clone_name`**: Generates a unique name for the next clone based on the group's `next_id`.
* **`register_clone`**: Adds a newly created clone to the group.
* **`remove_clone`**: Removes a clone from the group and updates the next clone ID.
* **`update_vm_metrics`**: Calculates and updates CPU usage metrics for a specific VM (primary or clone) in the group.
* **`evaluate_scaling`**: Calculates the group-wide average CPU load, updates consecutive tick counters, and returns the scaling status.
* **`sync_group`**: Synchronizes the group state against active VMs, pruning dead clones and reporting if the primary has died.

### DIFF
--- a/autoscaler.ml
+++ b/autoscaler.ml
@@ -142,17 +142,22 @@ module Cluster_manager = struct
       g.next_id <- max g.next_id (clone_id + 1)
     end
 
-  let register_clone ~user_name clone =
+  let register_clone user_name clone =
     match extract_name_and_clone_id (fst clone) with
-    | None -> Error (Fmt.str "Invalid clone name '%s'." (fst clone))
-    | Some (primary_name, clone_id) -> (
+    | Some (primary_name, _) -> (
         match find_group_by_name ~user_name ~unikernel_name:primary_name with
         | Some g ->
-            add_clone_to_group g clone clone_id;
+            g.next_id <- g.next_id + 1;
+            let new_name = Fmt.str "%s-clone-%d" primary_name g.next_id in
+            add_clone_to_group g (new_name, snd clone) g.next_id;
             g.last_scale_action <- Some (Mirage_ptime.now ());
-            Ok ()
+            Ok new_name
         | None ->
             Error (Fmt.str "No primary group found for '%s'." primary_name))
+    | None ->
+        let g = get_or_create ~user_name clone in
+        g.next_id <- g.next_id + 1;
+        Ok (fst clone)
 
   let find_or_create_group ~user_name ~unikernel_name t =
     match extract_name_and_clone_id unikernel_name with
@@ -172,26 +177,6 @@ module Cluster_manager = struct
         | None ->
             let g = get_or_create ~user_name (unikernel_name, t) in
             Ok g)
-
-  let next_clone_name user_name vm =
-    match extract_name_and_clone_id (fst vm) with
-    | Some (primary_name, _) -> (
-        match find_group_by_name ~user_name ~unikernel_name:primary_name with
-        | Some g ->
-            let id = g.next_id in
-            g.next_id <- g.next_id + 1;
-            Ok (Fmt.str "%s-clone-%d" primary_name id)
-        | None ->
-            Error
-              (Fmt.str
-                 "No primary group found for primary '%s' (derived from clone \
-                  '%s')."
-                 primary_name (fst vm)))
-    | None ->
-        let g = get_or_create ~user_name vm in
-        let id = g.next_id in
-        g.next_id <- g.next_id + 1;
-        Ok (Fmt.str "%s-clone-%d" (fst g.primary) id)
 
   let remove_clone ~user_name clone =
     match extract_name_and_clone_id (fst clone) with
--- a/autoscaler.ml
+++ b/autoscaler.ml
@@ -1,0 +1,326 @@
+(** how long to wait before checking the unikernel *)
+let poll_interval = Duration.of_min 5
+
+(** number of times a unikernel is checked before deciding if it's overloaded.
+*)
+let scale_up_trigger_ticks = 3
+
+(** if CPU usage >= 90% for [trigger_ticks] consercutive checks then this vm can
+    be cloned. *)
+let scale_up_threshold_percent = 90.0
+
+let scale_down_threshold_percent = 40.0
+let scale_down_trigger_ticks = 5
+
+(** after spawning a new vm, wait [cooldown_period] before checking that vm
+    again. *)
+let cooldown_period = 600.0
+
+(** if no stats are gotten for a particular vm after [death_timeout] then
+    consider the vm destroyed and prune it *)
+let death_timeout = 900.0
+
+module Cpu_monitor = struct
+  let rusage_to_float (sec, usec) =
+    let s = Int64.to_float sec in
+    let u = float_of_int usec in
+    s +. (u /. 1_000_000.0)
+
+  let get_total_cpu_time (r : Vmm_core.Stats.rusage) =
+    let user_t = rusage_to_float r.utime in
+    let sys_t = rusage_to_float r.stime in
+    user_t +. sys_t
+
+  type t = { last_cpu_time : float; last_wall_time : Ptime.t }
+
+  let create now (initial_rusage : Vmm_core.Stats.rusage) =
+    { last_cpu_time = get_total_cpu_time initial_rusage; last_wall_time = now }
+
+  let measure t now (current_rusage : Vmm_core.Stats.rusage) =
+    let curr_cpu_time = get_total_cpu_time current_rusage in
+    let curr_wall_time = now in
+    let cpu_delta = curr_cpu_time -. t.last_cpu_time in
+    let wall_span = Ptime.diff curr_wall_time t.last_wall_time in
+    let wall_delta = Ptime.Span.to_float_s wall_span in
+    { last_cpu_time = curr_cpu_time; last_wall_time = curr_wall_time }
+    |> fun t ->
+    if wall_delta <= 0.000001 then 0.0
+    else if cpu_delta < 0.0 then 0.0
+    else
+      let pct = cpu_delta /. wall_delta *. 100.0 in
+      Float.min 100.0 pct
+end
+
+type t = {
+  mutable monitor : Cpu_monitor.t;
+  mutable last_cpu_usage : float;
+  mutable last_stats_received : Ptime.t;
+}
+
+type status =
+  | Overloaded of
+      t (* the usage of the clone or vm that is checked in the group *)
+  | Pending of int * t
+    (* We have a high load but waiting for required number of polls to complete*)
+  | Pending_down of int * t
+  | Underloaded of string * t
+  | Cooldown of t
+  | Normal of t
+
+let create now initial_rusage =
+  {
+    monitor = Cpu_monitor.create now initial_rusage;
+    last_cpu_usage =
+      Cpu_monitor.measure
+        (Cpu_monitor.create now initial_rusage)
+        now initial_rusage;
+    last_stats_received = now;
+  }
+
+module Cluster_manager = struct
+  type group = {
+    primary : string * t;
+    mutable clones : (string * t) list;
+    mutable last_scale_action : Ptime.t option;
+    mutable last_stats_received : Ptime.t;
+    mutable next_id : int;
+    mutable consecutive_high_ticks : int;
+    mutable consecutive_low_ticks : int;
+  }
+
+  let clusters : (string, group) Hashtbl.t = Hashtbl.create 10
+
+  let in_cooldown now = function
+    | { last_scale_action = None; _ } -> false
+    | { last_scale_action = Some t; _ } ->
+        let span = Ptime.diff now t in
+        Ptime.Span.to_float_s span < cooldown_period
+
+  let get_or_create primary =
+    match Hashtbl.find_opt clusters (fst primary) with
+    | Some g -> g
+    | None ->
+        let g =
+          {
+            primary;
+            clones = [];
+            consecutive_high_ticks = 0;
+            consecutive_low_ticks = 0;
+            last_scale_action = None;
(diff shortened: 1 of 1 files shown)


autoscaler.ml
```diff
@@ -0,0 +1,296 @@
+let a_logs = Logs.Src.create "autoscaling-logs"
+let poll_interval = Duration.of_min 5
```
hannesm, line 4: maybe this should have a docstring as well?

autoscaler.ml
```diff
@@ -0,0 +1,296 @@
...
+
+(** number of times a cluster is checked before deciding if it's overloaded. *)
+let scale_up_trigger_ticks = 3
+
+(** if CPU usage >= 90% for [scale_up_trigger_ticks] consecutive checks then
+    this vm can be cloned. *)
+let scale_up_threshold_percent = 90.0
+
+let scale_down_threshold_percent = 40.0
+let scale_down_trigger_ticks = 5
```
hannesm, line 21: same as above, some documentation would be nice.

autoscaler.ml
```diff
@@ -0,0 +1,296 @@
...
+
+(** after spawning a new vm, wait [cooldown_period] before checking that vm
+    again. *)
+let cooldown_period = 600.0
+
+(** if no stats are gotten for a particular vm after [death_timeout] then
+    consider the vm destroyed and prune it *)
+let death_timeout = 900.0
```
hannesm: I wonder whether these should instead be Ptime.Span.t ?

autoscaler.ml
```diff
@@ -0,0 +1,296 @@
...
+    let elasped_time_in_seconds =
+      Ptime.Span.to_float_s elasped_time_difference
+    in
+    if elasped_time_in_seconds <= 0.000001 then 0.0
+    else if cpu_delta < 0.0 then 0.0
+    else
+      let pct = cpu_delta /. elasped_time_in_seconds *. 100.0 in
+      (* TODO: use numcpus to cap it at 100.0% if the vm has more than 1 cpu. Now most
+         vms use 1 cpu, so capping at 100% is fine. *)
+      Float.min 100.0 pct
```
hannesm: I don't quite understand why floating point values are used all over here, and what would be wrong to use microseconds as int instead. But I guess we have other things to do than to argue about that.

autoscaler.ml
```diff
@@ -0,0 +1,296 @@
...
+    { last_cpu_time = get_total_cpu_time initial_rusage; last_wall_time = now }
+
+  let measure t now (current_rusage : Vmm_core.Stats.rusage) =
+    let curr_cpu_time = get_total_cpu_time current_rusage in
+    let cpu_delta = curr_cpu_time -. t.last_cpu_time in
+    let elasped_time_difference = Ptime.diff now t.last_wall_time in
+    let elasped_time_in_seconds =
+      Ptime.Span.to_float_s elasped_time_difference
+    in
+    if elasped_time_in_seconds <= 0.000001 then 0.0
```
hannesm: I've no clue why this line is here, what is wrong when the computation below gets a small "elasped_time_in_seconds"?

Also, there's a typo -- it should be `ela_ps_ed`.

autoscaler.ml
```diff
@@ -0,0 +1,296 @@
...
+    else if cpu_delta < 0.0 then 0.0
```
hannesm: how can this happen?

autoscaler.ml
```diff
@@ -0,0 +1,296 @@
...
+      (* TODO: use numcpus to cap it at 100.0% if the vm has more than 1 cpu. Now most
+         vms use 1 cpu, so capping at 100% is fine. *)
+      Float.min 100.0 pct
+end
+
+type t = {
+  mutable monitor : Cpu_monitor.t;
+  mutable last_cpu_usage : float;
+  mutable last_stats_received : Ptime.t;
+}
```
hannesm, line 65: A type with all-mutable fields... this smells a bit... could this instead be pure immutable values, and you pass in/out a `t` (i.e. always construct a fresh one when you want to modify a field)?

autoscaler.ml
```diff
@@ -0,0 +1,296 @@
...
+  }
+
+module Cluster_manager = struct
+  type vm = string * t
+
+  type group = {
+    primary : vm;
+    mutable clones : vm list;
+    mutable last_scale_action : Ptime.t option;
+    mutable last_tick_update : Ptime.t option;
```
hannesm: is this good to be `option`? Or would just a `Ptime.t` be fine? (and then initialize to Ptime.epoch)

autoscaler.ml
```diff
@@ -0,0 +1,296 @@
...
+    | { last_scale_action = Some t; _ } ->
+        let span = Ptime.diff now t in
+        Ptime.Span.to_float_s span < cooldown_period
+
+  let should_tick now group =
+    match group.last_tick_update with
+    | None -> true
+    | Some t ->
+        let span = Ptime.diff now t in
+        Ptime.Span.to_float_s span >= Duration.to_f poll_interval
```
hannesm: Would this be possible to use Ptime.Span.t on both sides, and then avoid the generic comparison (and instead `Ptime.Span.compare`)

autoscaler.ml
```diff
@@ -0,0 +1,296 @@
...
+      g.clones <- clone :: g.clones;
+      g.next_id <- max g.next_id (clone_id + 1)
+    end
+
+  let register_clone user_name clone =
+    match extract_name_and_clone_id (fst clone) with
+    | Some (primary_name, _) -> (
+        match find_group_by_name ~user_name ~unikernel_name:primary_name with
+        | Some g ->
+            g.next_id <- g.next_id + 1;
```
hannesm: so here you increment the next_id by 1 for the clone...

autoscaler.ml
```diff
@@ -0,0 +1,296 @@
...
+            let new_name = Fmt.str "%s-clone-%d" primary_name g.next_id in
+            add_clone_to_group g (new_name, snd clone) g.next_id;
```
hannesm: and here, next_id is incremented again... is this next_id meant to be the next usable (free) id (so you'd increment after usage), or the last used id (when you'd need to increment before usage)?

autoscaler.ml
```diff
@@ -0,0 +1,296 @@
...
+            g.last_scale_action <- Some (Mirage_ptime.now ());
+            Ok new_name
+        | None ->
+            Error (Fmt.str "No primary group found for '%s'." primary_name))
+    | None ->
+        let g = get_or_create ~user_name clone in
+        g.next_id <- g.next_id + 1;
```
hannesm: same question as above.

autoscaler.ml
```diff
@@ -0,0 +1,296 @@
...
+                  "[Cluster Manager] No primary group found during removal for \
+                   primary '%s' (derived from clone '%s')."
+                  primary_name (fst clone));
+            Error "Primary group not found during removal")
+    | None ->
+        Logs.debug ~src:a_logs (fun m ->
+            m "[Cluster Manager] Invalid clone name '%s'." (fst clone));
+        Error "Invalid clone name"
+
+  let check_group_average group key now rusage =
```
hannesm: Would you mind to elaborate what this function is supposed to do, and what are the input arguments?

I had the impression from the name, it should compute the average CPU usage!?

But then, what does the `key` argument do? And why is there a `if String.equal ...`? What is the case for your group.primary :: group.clones that any element of this list is not name = key?

autoscaler.ml
```diff
@@ -0,0 +1,296 @@
...
+          trigger (Underloaded (clone_to_kill, current_vm_state))
+        else if is_high then
+          Ok (Pending (`Spawn, group.consecutive_high_ticks, current_vm_state))
+        else if is_low then
+          Ok (Pending (`Prune, group.consecutive_low_ticks, current_vm_state))
+        else Ok (Normal current_vm_state)
+
+  let is_dead now (vm : t) =
+    Ptime.Span.to_float_s (Ptime.diff now vm.last_stats_received)
+    > death_timeout
```
hannesm: again, would it be possible to use Ptime.Span.compare?

autoscaler.ml
```diff
@@ -0,0 +1,296 @@
...
+    in
+    match !current_vm_state with
+    | None -> Error "Current VM not found in group during average check"
+    | Some state ->
+        let average_usage =
+          total_usage /. float_of_int (List.length all_instances)
+        in
+        Ok (average_usage, state)
+
+  let check_group_status group key now rusage =
```
hannesm: I've no clue what `check_group_status` is supposed to do with the input arguments that exceed the `group`? Why is there a `key` and a `now` and a `rusage`?

autoscaler.ml
```diff
@@ -0,0 +1,296 @@
...
+                acc
+          end)
+        clusters []
+    in
+    List.iter
+      (fun key ->
+        Logs.debug ~src:a_logs (fun m ->
+            m "[Cluster Manager] Pruning dead cluster: %s" key);
+        Hashtbl.remove clusters key)
+      dead_keys
```
hannesm: As discussed on matrix, I'm not a big fan of Hashtbl and background tasks. Can we design something that is a bit more robust and doesn't rely on "prune_dead_clusters"?

I don't quite understand the semantics. What should happen if a unikernel (the primary) disappears, which has been scaled up? It looks like it is then removed from the `clusters` hash table, but what happens with all the clones? Won't they be re-inserted? How's the code dealing with a "group" that doesn't have a "primary"?

autoscaler.ml
```diff
@@ -0,0 +1,313 @@
...
+
+  let timeval_to_float (sec, usec) =
+    let s = Int64.to_float sec in
+    let u = float_of_int usec in
+    s +. (u /. 1_000_000.0)
+
+  let get_total_cpu_time (r : Vmm_core.Stats.rusage) =
+    let user_t = timeval_to_float r.utime in
+    let sys_t = timeval_to_float r.stime in
+    user_t +. sys_t
```
hannesm: there's as well a `runtime` field in `kinfo_mem`. any reason why rusage is used here? (I'm curious, there's no need to change any code.)

PizieDust: so stats is `type t = rusage * kinfo_mem option * vmm option * ifdata list`, since `kinfo_mem` is optional, I decided to use rusage since i was sure it's always present

autoscaler.ml
hannesm: thanks for your comment. I'm still wondering what the actions should be and triggered when.

so, for a new measurement that arrives certainly we want to compute the cpu load.

now, the iterating over all clones and the primary, this should be done once when all measurements arrived, or? so, shouldn't 1 and 2-4 be separate? maybe once the primary measurement is received is the time when to compute the group average?

hannesm:
I think this is fine,still I am missing some more high-level view of what should happen and what is being computed when and what is being kept in memory...

autoscaler.ml
PizieDust: in the case that a restart happens, this computation can be negative;
say we had a last value of 120.0 and a restart happens, and we get a new value 0.5, then this will return 0.5 - 120.0

autoscaler.ml
PizieDust: I added a comment for it now

autoscaler.ml
PizieDust: In the latest commits, we should compute the group averate now only when we receive stats for the primary.

autoscaler.ml
PizieDust: so it was mainly becuase `Ptime.Span.to_float_s` gives me a `float`, but `Ptime.Span.to_int_s` gives an `int option`, but i removed all usage of floats :)

PizieDust: https://github.com/robur-coop/mollymawk/pull/250/commits/24ba643c2aad7fb04f8bbda347800ef8032fd4a3

PizieDust:
> ...I am missing some more high-level view of what should happen 

we need to receive metrics from albatross. now we process metrics for clones and the primary vm individually to calculate their cpu load, but when its the primary vm, then we do a group-wide average calculation and update tick counters. then depending on the average cpu load, we can either scale up, scale down, or do nothing.

> and what is being computed when 

- when any vm stats arrive: individual CPU load is computed
- when the primary cm stats arrive: the cluster average load is computed. overload/underload counters tick up or reset based on this average.

>and what is being kept in memory...

- Group State:
  - the list of running vms in the group (primary + clones)
  - the last scale action timestamp 
  - tick counters (consecutive_high_ticks and consecutive_low_ticks)
  - the next clone index for naming the new clone
- Per-VM State:
  - last computed cpu percentage.
  - baseline cumulative cpu time and wall-clock timestamp (to calculate the next delta).
  - last time stats were received (for death detection).

Issue #68, opened 2024-10-15
Data structures and persistency for user data


hannesm:
At the moment, we use a list. And then we iterate over it for everything we need.

We also store everything on disk. With the rise of CSRF tokens / cookies, we may reconsider and store only some things on disk.

Also, these cookies seam to be never deleted, maybe we should remove them once used (for CSRF), and when expired.

Maybe a Map data structure, or multiple maps, are useful & sensible. They may be constructed when data is read at startup, and extended when data is stored onto disk (where the on-disk format is still a list).

hannesm:
Preparational work is done in #89 -- though this didn't change the data structures nor the data on disk.

PizieDust:
see as well https://github.com/robur-coop/mollymawk/pull/251#discussion_r3498286414

PR #251, opened 2026-05-30
handle metrics received

PizieDust:
This PR connects the metrics reeived from albatross with the autoscaling logic to determine if we should spawn new vms or destroy clones.

### DIFF
--- a/albatross.ml
+++ b/albatross.ml
@@ -614,11 +614,14 @@ module Make (S : Tcpip.Stack.V4V6) = struct
         raw_query stack t ~name:vmm_name certificates cmd
           (block_data vmm_name f)
 
-  let query_stats stack t ~domain f =
+  let query_stats stack t f =
     let cmd = `Stats_cmd `Stats_subscribe in
-    match certs t domain dot_name cmd with
-    | Error str -> Lwt.return (Error str)
-    | Ok (vmm_name, certificates) ->
+    let vmm_name = Vmm_core.Name.root in
+    match gen_cert t cmd dot_name with
+    | Error err ->
+        t.status <- Status.update t.status (Status.make `Certificate err);
+        Lwt.return (Error err)
+    | Ok certificates ->
         raw_query stack t ~name:vmm_name certificates cmd
           (stats_data vmm_name f)
 
--- a/albatross.ml
+++ b/albatross.ml
@@ -622,6 +622,7 @@ module Make (S : Tcpip.Stack.V4V6) = struct
         t.status <- Status.update t.status (Status.make `Certificate err);
         Lwt.return (Error err)
     | Ok certificates ->
+        set_online t;
         raw_query stack t ~name:vmm_name certificates cmd
           (stats_data vmm_name f)
 
--- a/unikernel.ml
+++ b/unikernel.ml
@@ -3245,71 +3245,41 @@ struct
     Lwt.async loop
 
   let start_background_scaler_scheduler stack store albatross_instances_ref =
-    let active_streams = Hashtbl.create 5 in
-    let spawn_stats_stream (user : User_model.user) instance_name instance =
-      let user_name = user.User_model.name in
+    let active_streams = ref Map.empty in
+    let spawn_stats_stream instance =
       let rec stream_loop () =
         Lwt.catch
           (fun () ->
-            unikernels_stats stack instance user >>= function
+            unikernels_stats stack instance store >>= function
             | Ok () -> Mirage_sleep.ns (Duration.of_sec 10) >>= stream_loop
             | Error err ->
                 Logs.debug ~src:stats_src (fun m ->
-                    m "Stats stream for %s on %s failed: %s"
-                      (Configuration.name_to_str user_name)
-                      (Configuration.name_to_str instance_name)
+                    m "Stats stream for %s failed: %s"
+                      (Configuration.name_to_str instance.configuration.name)
                       err);
                 Mirage_sleep.ns (Duration.of_sec 10) >>= stream_loop)
-          (function
-            | Lwt.Canceled ->
-                Logs.debug ~src:stats_src (fun m ->
-                    m "Stats stream for %s on %s cancelled (user deleted)."
-                      (Configuration.name_to_str user_name)
-                      (Configuration.name_to_str instance_name));
-                Lwt.return_unit
-            | exn ->
-                Logs.info ~src:stats_src (fun m ->
-                    m "Exception in stats stream for %s on %s: %s"
-                      (Configuration.name_to_str user_name)
-                      (Configuration.name_to_str instance_name)
-                      (Printexc.to_string exn));
-                Hashtbl.remove active_streams (user_name, instance_name);
-                Lwt.return_unit)
+          (fun exn ->
+            Logs.err ~src:stats_src (fun m ->
+                m "Exception in stats stream for %s: %s. Retrying in 10s..."
+                  (Configuration.name_to_str instance.configuration.name)
+                  (Printexc.to_string exn));
+            Mirage_sleep.ns (Duration.of_sec 10) >>= stream_loop)
       in
       stream_loop ()
     in
     let rec loop () =
       Lwt.pause () >>= fun () ->
-      let current_users = Store.users store in
       let current_instances = !albatross_instances_ref in
-      let valid_keys = Hashtbl.create 5 in
-      List.iter
-        (fun user ->
-          Albatross.Albatross_map.iter
-            (fun instance_name instance ->
-              let key = (user.User_model.name, instance_name) in
-              Hashtbl.replace valid_keys key ();
-              if not (Hashtbl.mem active_streams key) then begin
-                Logs.debug ~src:stats_src (fun m ->
-                    m "Spawning new stats stream for %s on %s"
-                      (Configuration.name_to_str user.User_model.name)
-                      (Configuration.name_to_str instance_name));
-                let p = spawn_stats_stream user instance_name instance in
-                Hashtbl.replace active_streams key p
-              end)
-            current_instances)
-        current_users;
-      let to_remove =
-        Hashtbl.fold
-          (fun key p acc ->
-            if not (Hashtbl.mem valid_keys key) then begin
-              Lwt.cancel p;
-              key :: acc
-            end
-            else acc)
-          active_streams []
-      in
-      List.iter (fun key -> Hashtbl.remove active_streams key) to_remove;
+      Albatross.Albatross_map.iter
+        (fun instance_name instance ->
+          let key = Configuration.name_to_str instance_name in
+          if not (Map.mem key !active_streams) then begin
+            Logs.debug ~src:stats_src (fun m ->
+                m "Spawning new stats stream for albatross instance %s" key);
+            let p = spawn_stats_stream instance in
+            active_streams := Map.add key p !active_streams
+          end)
+        current_instances;
       Mirage_sleep.ns (Duration.of_sec 30) >>= loop
     in
     Lwt.async loop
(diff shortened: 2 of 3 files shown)


unikernel.ml
```diff
@@ -3050,6 +3028,131 @@ struct
...
+                   unikernel_name err))
+    else Lwt.return (Ok ())
+
+  let unikernels_stats stack instance (user : User_model.user) =
+    let user_name = user.User_model.name in
+    let cb name st =
+      let unikernel_name =
+        match Vmm_core.Name.name name with
+        | Some label -> Vmm_core.Name.Label.to_string label
+        | None -> Vmm_core.Name.to_string name
```
hannesm: ? When is this supposed to happen, and why is using "Vmm_core.Name.to_string name" the right thing here?

unikernel.ml
```diff
@@ -3050,6 +3028,131 @@ struct
...
+      in
+      Logs.debug ~src:stats_src (fun m ->
+          m "[Stats] Received stats for %s as %a" unikernel_name
+            Vmm_core.Stats.pp st);
+      Lwt.async (fun () ->
```
hannesm, line 3213: Why is this an Lwt.async?

unikernel.ml
```diff
@@ -3050,6 +3028,131 @@ struct
...
+          handle_stats stack instance st ~unikernel_name user >>= function
```
hannesm: why does handle_stats need a `stack` argument?

unikernel.ml
```diff
@@ -3160,11 +3263,12 @@ struct
 
   let start_background_scaler_scheduler stack store albatross_instances_ref =
```
hannesm, line 3340: I don't quite understand the complexity of this function.

How often and when are "stats_stream" spawned? From my understanding, mollymawk would need _one_ stream for each albatross instance, but it looks like for each user and each albatross there's eventually a stream?

unikernel.ml
hannesm: I don't quite understand why there's need for yet another Hash table, "valid_keys".

unikernel.ml
PizieDust: when we get metrics, we want to know for which unikernel the metrics are for, and from the name, we can know who the unikernel belongs to (user) and we can determine if it's a clone or a primary.

when it's just a clone, we then just update the metrics. 
when it's a primary, we evaluate if to scale or not

we no longer use Vmm_core.Name.to_string, now we return an error

unikernel.ml
PizieDust: we have two loops:
- loop which runs every 30 seconds and starts a stream for each albatross instance. it has a map active_streams it maintains. 
- stream_loop which runs every 10 seconds and manages stats for each albatross instance

in the new design, mollymawk uses 1 stream for an instance.

unikernel.ml
PizieDust: we now use a map

albatross.ml
```diff
@@ -614,11 +614,14 @@ module Make (S : Tcpip.Stack.V4V6) = struct
...
-    | Error str -> Lwt.return (Error str)
-    | Ok (vmm_name, certificates) ->
+    let vmm_name = Vmm_core.Name.root in
+    match gen_cert t cmd dot_name with
+    | Error err ->
+        t.status <- Status.update t.status (Status.make `Certificate err);
+        Lwt.return (Error err)
+    | Ok certificates ->
         raw_query stack t ~name:vmm_name certificates cmd
           (stats_data vmm_name f)
```
hannesm, line 627: should we here do a set_online?

unikernel.ml
```diff
@@ -3033,6 +3011,175 @@ struct
               ~data:(`String (Fmt.str "Test failed: %s" (String.escaped err)))
               `Bad_request)
 
+  let scaling_groups = ref Map.empty
+  let stats_src = Autoscaler.a_logs
```
hannesm: you could add
```OCaml
module Log = (val Logs.src_log stats_src : Logs.LOG)
```

and then use below `Log.info` etc.

unikernel.ml
```diff
@@ -3033,6 +3011,175 @@ struct
...
+
+  let evaluate_scaling group now key =
+    match Autoscaler.Cluster_manager.evaluate_scaling group now with
+    | Error err ->
+        Lwt.return_error
+          (Fmt.str "Error evaluating scaling for cluster %s: %s" key err)
+    | Ok (status, group) -> (
+        scaling_groups := Map.add key group !scaling_groups;
+        let log msg = Logs.info ~src:stats_src (fun m -> m "%s" msg) in
```
hannesm: I don't like this local binding:
- it makes the code less clear (Log.info ...) is much easier to comprehend than "look for `log` and find what it is actually doing"
- at use site, you use `Fmt.str` now, and the `m "%s" msg` does another formatting round -- so it allocates more and uses more resources.

See above how you could use `Log.info` directly below.

unikernel.ml
```diff
@@ -3033,6 +3011,175 @@ struct
...
+                  (Fmt.str "User %s not found."
+                     (Configuration.name_to_str user_label))
+            | Some user ->
+                let primary_name =
+                  match
+                    Autoscaler.Cluster_manager.extract_name_and_clone_id
+                      unikernel_name
+                  with
+                  | Some (p, _) -> p
+                  | None -> unikernel_name
```
hannesm: could as well be `Option.value ~default:unikernel_name (Option.map fst (Autoscaler...))`.

unikernel.ml
```diff
@@ -3033,6 +3011,175 @@ struct
...
+                in
+                let is_scaling_enabled =
+                  List.exists
+                    (fun (p : User_model.unikernel_scaling_policy) ->
+                      String.equal
+                        (Configuration.name_to_str p.name)
+                        primary_name)
```
hannesm: I'm curious, why use String here, instead of primary_name being a Vmm_core.Name.Label.t? I'm in favour of using types where we can, and avoid the conversions to string... esp since you already have `label` bound as a Vmm_core.Name.Label.t above.

unikernel.ml
```diff
@@ -3033,6 +3011,175 @@ struct
...
+                  Logs.debug ~src:stats_src (fun m ->
+                      m "[Stats] Received stats for %s as %a" unikernel_name
+                        Vmm_core.Stats.pp_rusage rusage);
+                  let now = Mirage_ptime.now () in
+                  let key =
+                    Configuration.name_to_str user.name ^ "-" ^ primary_name
+                  in
+                  (* Get the group this vm should belong to *)
+                  let group =
+                    match Map.find_opt key !scaling_groups with
```
hannesm: data structure layout: it seems you have a flat map (where key is a string) `scaling_groups`.. also, the code is taking care of "user" and "unikernel_name" -- so would a nested map (so, scaling_groups would have as key only the user, and then return another map which keys are the unikernel names) be more suitable? Also, you may use Vmm_core.Name.Label maps instead of string maps...

unikernel.ml
```diff
@@ -3033,6 +3011,175 @@ struct
...
+                    | Some g -> g
+                    | None ->
+                        let primary_vm =
+                          (primary_name, Autoscaler.create now rusage)
+                        in
+                        Autoscaler.Cluster_manager.create_group primary_vm
+                  in
+                  (* If the vm is a ghost clone, add it to the group*)
```
hannesm: I don't understand what a `ghost clone` would be, would you mind to explain this in the comment?

unikernel.ml
```diff
@@ -3033,6 +3011,176 @@ struct
...
+                       (Autoscaler.Cluster_manager.extract_name_and_clone_id
+                          unikernel_name))
+                in
+                let is_scaling_enabled =
+                  List.exists
+                    (fun (p : User_model.unikernel_scaling_policy) ->
+                      String.equal
+                        (Configuration.name_to_str p.name)
+                        primary_name)
+                    user.scaling_policies
```
hannesm, line 3140: It looks to me that the datatype of `user.scaling_policies` should eventually be a map as well, instead of a list?

Not sure where to raise this issue, since I've opened #68 a long time ago... we can of course fix this later...

unikernel.ml
```diff
@@ -3033,6 +3011,224 @@ struct
               ~data:(`String (Fmt.str "Test failed: %s" (String.escaped err)))
               `Bad_request)
 
+  module Label_map = Albatross.Albatross_map
+
+  let scaling_groups :
+      Autoscaler.Cluster_manager.group Label_map.t Label_map.t ref =
+    ref Label_map.empty
+
+  module Autoscaler_Log = (val Logs.src_log Autoscaler.a_logs : Logs.LOG)
```
hannesm: I usually use the module name `Log` here in various libraries. This is shorter. Certainly, then we need to distinguish `Log` from `Logs` -- but IMHO that's fine.

unikernel.ml
```diff
@@ -3142,72 +3338,42 @@ struct
...
-    let active_streams = Hashtbl.create 5 in
-    let spawn_stats_stream user_name instance_name instance =
+    let active_streams = ref Map.empty in
+    let spawn_stats_stream instance =
       let rec stream_loop () =
         Lwt.catch
           (fun () ->
-            unikernels_stats stack instance user_name >>= function
+            unikernels_stats stack instance store >>= function
             | Ok () -> Mirage_sleep.ns (Duration.of_sec 10) >>= stream_loop
```
hannesm: here, there's a recursive call to `stream_loop` from within the `Lwt.catch` (which sets up some exception handling) -- which will lead to a stack overflow after enough iterations.

It would be better to move the recursive call outside the Lwt.catch -- see e.g. https://github.com/mirage/mirage-net-solo5/blob/v0.8.1/src/netif.ml#L129-L149 for further explanations.

It may be sensible to mark the recursive call with `[@tailcall]` annotation (so `(stream_loop [@tailcall]) ()`) to avoid stack overflow issues. You can, as in mirage-net-solo5, return a result from the main stream_loop code to decide whether to do another recursive call or not.

hannesm:
I looked over this PR, it looks mostly good - the non-tail-recursive-function should be fixed. Apart from that, good to merge (some questions/issue stay unsolved, but I'm as well out of energy for reviewing).

unikernel.ml
PizieDust: thanks. I was trying to avoid confusing it with Log but I've used the short Log version now :)

unikernel.ml
PizieDust: thanks for catching that.

unikernel.ml
PizieDust: we can change the datatypes in a dedicated PR. Since #68 is open, I think we don't need to open another issue. I'll add your comment there as well

unikernel.ml
PizieDust: here i want handle_stats to be asynchronous even though the callback to  Albatross_state.query_stats is synchronous. so to be sure we have no blocking operations this is why i wrapped it in an Lwt.async

hannesm: but why? I still don't understand -- so you're spawning in `start_background_scaler_scheduler` an asynchronous thing, and there you call recursively `stream_loop` which (called for each albatross instance) does an `Lwt.async` for each received statistic.

This leads to a pretty high amount of Lwt tasks (which is a concern), and I still don't understand why they should be in an Lwt.async -- what is your `handle_stats` supposed to do? only update data/metrics/load in some internal memory; or do you as well want to create/destroy unikernels from that? I'd advise against the latter.

As mentioned earlier, I don't quite understand the design of the interaction of the components. https://github.com/robur-coop/mollymawk/pull/251#discussion_r3496834399

From how albatross interacts:
- mollymawk subscribes to the statistics subsystem (`` `Stats_cmd `Stats_subscribe ``)
- then albatross sends whenever it receives statistics those to the client (`` `Data `Stats_data _ ``)
- you should not need to setup new connections to albatross unless that stream of statistics is closed from the albatross side

I guess I'll have to dig a bit more into the mollymawk code to understand what it aims to do.

PR #253, opened 2026-05-30
Periodically prune dead clones and clusters

PizieDust:
While autoscaling is happening, we should clear the hashtable periodically so it doesnt grow indefinitely

### DIFF
--- a/unikernel.ml
+++ b/unikernel.ml
@@ -3384,6 +3384,25 @@ struct
     in
     Lwt.async loop
 
+  let run_prune_dead_clusters () =
+    let rec loop () =
+      Logs.info (fun m ->
+          m "Pruning dead clusters in %.0f seconds" Autoscaler.death_timeout);
+      Mirage_sleep.ns (Duration.of_f Autoscaler.death_timeout) >>= fun () ->
+      Lwt.catch
+        (fun () ->
+          Logs.info (fun m -> m "Starting background pruning...");
+          Lwt.return
+            (Autoscaler.Cluster_manager.prune_dead_clusters
+               (Mirage_ptime.now ())))
+        (fun exn ->
+          Logs.err (fun m ->
+              m "Background pruning failed: %s" (Printexc.to_string exn));
+          Lwt.return_unit)
+      >>= fun () -> loop ()
+    in
+    Lwt.async loop
+
   let request_handler stack management_happy_eyeballs management_domain
       albatross_instances js_file css_file imgs grafana_file store http_client
       happy_eyeballs flow (_ipaddr, _port) reqd =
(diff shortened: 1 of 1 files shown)


PizieDust:
The implementation of this functionality may not be needed if https://github.com/robur-coop/albatross/issues/278 is implemented.

unikernel.ml
```diff
@@ -3384,6 +3384,65 @@ struct
...
     Lwt.async loop
 
+  let run_prune_dead_clusters stack albatross_instances =
+    let rec loop () =
+      Logs.info (fun m ->
+          m "Pruning dead clusters in %a" Ptime.Span.pp Autoscaler.death_timeout);
+      let sleep_ns =
+        match Ptime.Span.to_int_s Autoscaler.death_timeout with
+        | None -> Duration.of_sec 600
+        | Some s -> Duration.of_sec s
```
hannesm: could as well be:
```OCaml
Duration.of_sec (Option.value ~default:600 (Ptime.Span.to_int_s Autoscaler.death_timeout))
```

unikernel.ml
```diff
@@ -3384,6 +3384,65 @@ struct
     in
     Lwt.async loop
 
+  let run_prune_dead_clusters stack albatross_instances =
+    let rec loop () =
+      Logs.info (fun m ->
+          m "Pruning dead clusters in %a" Ptime.Span.pp Autoscaler.death_timeout);
```
hannesm: not sure about the log level, is this intentional on the info level?

unikernel.ml
```diff
@@ -3384,6 +3384,65 @@ struct
...
+      let sleep_ns =
+        match Ptime.Span.to_int_s Autoscaler.death_timeout with
+        | None -> Duration.of_sec 600
+        | Some s -> Duration.of_sec s
+      in
+      Mirage_sleep.ns sleep_ns >>= fun () ->
+      Lwt.catch
+        (fun () ->
+          Logs.info (fun m -> m "Starting background pruning...");
```
hannesm: same as above, is this good on the info level?

unikernel.ml
```diff
@@ -3384,6 +3384,65 @@ struct
...
+                          Label_map.add user_name updated_unikernel_map
+                            !scaling_groups)
+                unikernel_map))
+        (fun exn ->
+          Logs.err (fun m ->
+              m "Background pruning failed: %s" (Printexc.to_string exn));
+          Lwt.return_unit)
+      >>= fun () -> loop ()
+    in
+    Lwt.async loop
```
hannesm, line 3449: similar to the email notifications, you run the background task without any timeout -- so when this gets stuck, it is not run anymore. I recommend to use a `Lwt.join` with a timeout, similar to https://git.robur.coop/robur/opam-mirror/pulls/50

PizieDust: Thank you. this helps a great deal

PR #254, opened 2026-05-31
Spawn a new clone

PizieDust:
In this PR, when it has been determined by the autoscaling module that a vm is overloaded, we spawn a new clone. 
- We first get the configuration of the primary vm,
- Then we check if it has any block devices. At the moment we can't scale vms which have a block device, as block devices can't be shared by multiple unikernels.
- If the unikernel has no block device, then we construct a config for the new clone:
     - we remove the mac addresses of the bridges, so the clone is issued new mac addresses
     - we assume we have DNSvizor running on any network interfaces so IP's get assigned
     - we add to the arguments, `--name=<primary_name> so DNSvizor knows to register the ip correctly

### DIFF
--- a/albatross.ml
+++ b/albatross.ml
@@ -365,6 +365,23 @@ module Make (S : Tcpip.Stack.V4V6) = struct
     in
     continue_reading name dec d tls_flow
 
+  let get_unikernel_binary name f tls_flow d =
+    let dec = function
+      | Error s ->
+          Logs.err (fun m ->
+              m "albatross stop reading unikernel binary %a: error %s"
+                Vmm_core.Name.pp name s);
+          Error ()
+      | Ok (hdr, `Success (`Unikernel_image (_compressed, data))) -> f data
+      | Ok w ->
+          Logs.warn (fun m ->
+              m "albatross unexpected reply, need unikernel binary, got %a"
+                (Vmm_commands.pp_wire ~verbose:false)
+                w);
+          Ok ()
+    in
+    continue_reading name dec d tls_flow
+
   let raw_query (stack : S.t) t ?(name = Vmm_core.Name.root) certificates cmd
       ?push f =
     let open Lwt.Infix in
@@ -613,6 +630,14 @@ module Make (S : Tcpip.Stack.V4V6) = struct
     | Ok (vmm_name, certificates) ->
         raw_query stack t ~name:vmm_name certificates cmd
           (block_data vmm_name f)
+
+  let query_unikernel_binary stack t ~domain ~name f =
+    let cmd = `Unikernel_cmd (`Unikernel_get 4) in
+    match certs t domain name cmd with
+    | Error str -> Lwt.return (Error str)
+    | Ok (vmm_name, certificates) ->
+        raw_query stack t ~name:vmm_name certificates cmd
+          (get_unikernel_binary vmm_name f)
 
   let query_stats stack t f =
     let cmd = `Stats_cmd `Stats_subscribe in
--- a/albatross.ml
+++ b/albatross.ml
@@ -263,19 +263,31 @@ module Make (S : Tcpip.Stack.V4V6) = struct
               m "albatross buffer too small: %u bytes, requires %u bytes"
                 (String.length data - 4 - off)
                 len);
-          acc)
-      else if String.length data = off then acc
+          (acc, None))
+      else if String.length data = off then (acc, None)
       else (
         Logs.warn (fun m ->
             m "albatross buffer too small: %u bytes leftover"
               (String.length data - off));
-        acc)
+        (acc, None))
     in
-    split [] data 0 |> List.rev
+    split [] data 0 |> fun (acc, left) -> (List.rev acc, left)
 
-  let rec continue_reading name f d flow =
+  let split_many_with_leftover data =
+    let rec split acc data off =
+      if String.length data - off >= 4 then
+        let len = Int32.to_int (String.get_int32_be data off) in
+        if String.length data - off >= 4 + len then
+          split (String.sub data (4 + off) len :: acc) data (off + len + 4)
+        else (acc, Some (String.sub data off (String.length data - off)))
+      else if String.length data = off then (acc, None)
+      else (acc, Some (String.sub data off (String.length data - off)))
+    in
+    split [] data 0
+
+  let rec continue_reading name f ?(split = split_many) d flow =
     let open Lwt.Infix in
-    let bufs = split_many d in
+    let bufs, left_over = split d in
     let r =
       List.fold_left
         (fun acc d ->
@@ -287,7 +299,13 @@ module Make (S : Tcpip.Stack.V4V6) = struct
     | Ok () -> (
         TLS.read flow >>= fun r ->
         match r with
-        | Ok (`Data d) -> continue_reading name f (Cstruct.to_string d) flow
+        | Ok (`Data d) -> (
+            match left_over with
+            | None -> continue_reading name f (Cstruct.to_string d) flow
+            | Some left_over_string ->
+                continue_reading name f ~split:split_many_with_leftover
+                  (left_over_string ^ Cstruct.to_string d)
+                  flow)
         | Ok `Eof ->
             Logs.info (fun m ->
                 m "albatross received eof while reading stream %a"
@@ -372,7 +390,9 @@ module Make (S : Tcpip.Stack.V4V6) = struct
               m "albatross stop reading unikernel binary %a: error %s"
                 Vmm_core.Name.pp name s);
           Error ()
-      | Ok (hdr, `Success (`Unikernel_image (_compressed, data))) -> f data
+      | Ok (hdr, `Success (`Unikernel_image (_compressed, data))) ->
+          let _ = f data in
+          Error ()
       | Ok w ->
           Logs.warn (fun m ->
               m "albatross unexpected reply, need unikernel binary, got %a"
@@ -380,7 +400,7 @@ module Make (S : Tcpip.Stack.V4V6) = struct
                 w);
           Ok ()
     in
-    continue_reading name dec d tls_flow
+    continue_reading ~split:split_many_with_leftover name dec d tls_flow
 
   let raw_query (stack : S.t) t ?(name = Vmm_core.Name.root) certificates cmd
       ?push f =
--- a/albatross.ml
+++ b/albatross.ml
@@ -626,43 +626,6 @@ module Make (S : Tcpip.Stack.V4V6) = struct
         raw_query stack t ~name:vmm_name certificates cmd
           (stats_data vmm_name f)
 
-  let unikernel_get_data name f tls_flow d =
-    let open Lwt.Infix in
-    let rec read_loop acc =
-      if String.length acc >= 4 then
-        let len = Int32.to_int (String.get_int32_be acc 0) in
-        if String.length acc >= 4 + len then (
-          match decode_reply acc with
-          | Error msg -> Lwt.return_error msg
-          | Ok (_, `Success (`Unikernel_image (compressed, binary))) ->
-              f (compressed, binary)
-          | Ok w ->
-              Logs.warn (fun m ->
-                  m "albatross unexpected reply %a"
-                    (Vmm_commands.pp_wire ~verbose:false)
-                    w);
-              Lwt.return_error "Unexpected reply for Unikernel_get")
-        else
-          TLS.read tls_flow >>= function
-          | Ok (`Data d) -> read_loop (acc ^ Cstruct.to_string d)
-          | Ok `Eof -> Lwt.return_error "EOF while reading unikernel image"
-          | Error _ -> Lwt.return_error "TLS error"
-      else
-        TLS.read tls_flow >>= function
-        | Ok (`Data d) -> read_loop (acc ^ Cstruct.to_string d)
-        | Ok `Eof -> Lwt.return_error "EOF while reading unikernel image header"
-        | Error _ -> Lwt.return_error "TLS error"
-    in
-    read_loop d
-
-  let query_unikernel_get stack t ~domain ~name f =
-    let cmd = `Unikernel_cmd (`Unikernel_get 0) in
-    match certs t domain name cmd with
-    | Error str -> Lwt.return (Error str)
-    | Ok (vmm_name, certificates) ->
-        raw_query stack t ~name:vmm_name certificates cmd
-          (unikernel_get_data vmm_name f)
-
   let set_policy stack t ~domain policy =
     let open Lwt.Infix in
     let old_policies = t.policies in
--- a/albatross.ml
+++ b/albatross.ml
@@ -626,6 +626,43 @@ module Make (S : Tcpip.Stack.V4V6) = struct
         raw_query stack t ~name:vmm_name certificates cmd
           (stats_data vmm_name f)
 
+  let unikernel_get_data name f tls_flow d =
+    let open Lwt.Infix in
+    let rec read_loop acc =
+      if String.length acc >= 4 then
+        let len = Int32.to_int (String.get_int32_be acc 0) in
+        if String.length acc >= 4 + len then (
+          match decode_reply acc with
+          | Error msg -> Lwt.return_error msg
+          | Ok (_, `Success (`Unikernel_image (compressed, binary))) ->
+              f (compressed, binary)
+          | Ok w ->
+              Logs.warn (fun m ->
+                  m "albatross unexpected reply %a"
+                    (Vmm_commands.pp_wire ~verbose:false)
+                    w);
+              Lwt.return_error "Unexpected reply for Unikernel_get")
+        else
+          TLS.read tls_flow >>= function
+          | Ok (`Data d) -> read_loop (acc ^ Cstruct.to_string d)
+          | Ok `Eof -> Lwt.return_error "EOF while reading unikernel image"
+          | Error _ -> Lwt.return_error "TLS error"
+      else
+        TLS.read tls_flow >>= function
+        | Ok (`Data d) -> read_loop (acc ^ Cstruct.to_string d)
+        | Ok `Eof -> Lwt.return_error "EOF while reading unikernel image header"
+        | Error _ -> Lwt.return_error "TLS error"
+    in
+    read_loop d
+
+  let query_unikernel_get stack t ~domain ~name f =
+    let cmd = `Unikernel_cmd (`Unikernel_get 0) in
+    match certs t domain name cmd with
+    | Error str -> Lwt.return (Error str)
+    | Ok (vmm_name, certificates) ->
+        raw_query stack t ~name:vmm_name certificates cmd
+          (unikernel_get_data vmm_name f)
+
   let set_policy stack t ~domain policy =
     let open Lwt.Infix in
     let old_policies = t.policies in
--- a/autoscaler.ml
+++ b/autoscaler.ml
@@ -156,6 +156,9 @@ module Cluster_manager = struct
     else g
 
   let register_clone group clone =
+    (* TODO: add a validation rule during unikernel creation so users 
+            can't name any vm with the suffix "-clone-INT". 
+            What about unikernels deployed directly on albatross? *)
     let name = fst clone in
     match extract_name_and_clone_id name with
     | Some (primary_name, clone_id) ->
--- a/autoscaler.ml
+++ b/autoscaler.ml
@@ -205,6 +205,8 @@ module Cluster_manager = struct
           (Fmt.str "Clone name '%s' is not a valid format"
              (Configuration.name_to_str clone_name))
 
+  let remove_cooldown group = { group with last_scale_action = Ptime.epoch }
+
   (** [update_vm_metrics group key now rusage] calculates the CPU usage for the
       VM [key] using [rusage] at time [now]. It updates the cached state of VM
       [key] in the [group] (either primary or one of the clones). Returns the
--- a/unikernel.ml
+++ b/unikernel.ml
@@ -3019,6 +3019,21 @@ struct
 
   module Log = (val Logs.src_log Autoscaler.a_logs : Logs.LOG)
 
+  let least_used_cpuid stack albatross user_name =
+    user_unikernels_by_instance stack albatross user_name >|= fun unikernels ->
+    match Albatross_state.policy ~domain:user_name albatross with
+    | Error err -> Error err
+    | Ok None -> Error "No policy found for user"
+    | Ok (Some p) -> (
+        let cpu_usage_count = Utils.cpu_usage_count p unikernels in
+        let sorted_cpus =
+          cpu_usage_count
+          |> List.sort (fun (_, c1) (_, c2) -> Int.compare c1 c2)
+        in
+        match sorted_cpus with
+        | (id, _) :: _ -> Ok id
+        | [] -> Error "No cpuids available in user policy")
+
   let put_group ~(user_name : Vmm_core.Name.Label.t)
       ~(unikernel_name : Vmm_core.Name.Label.t) group =
     let unikernel_map =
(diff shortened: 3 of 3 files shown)


PizieDust:
I wonder if for the new clones we should include an additional `--name=<PRIMARY_NAME>` in the arguments so dnsvizor registers the IP to the same domain name

unikernel.ml
```diff
@@ -3030,6 +3030,102 @@ struct
 
   let stats_src = Logs.Src.create "albatross-stats"
 
+  let least_used_cpuid stack albatross user_name =
```
PizieDust, line 3022: this function is meant to return the least used cpuid for any albatross instance. eventually, we want to be able to scale to another albatross instance if the current instance is saturated.

unikernel.ml
```diff
@@ -3030,6 +3030,102 @@ struct
...
+    | Ok primary_vmm_name, Ok clone_vmm_name -> (
+        user_unikernel stack albatross ~user_name:user.name
+          ~unikernel_name:primary_vmm_name
+        >>= function
+        | Error err ->
+            Lwt.return_error
+              (Fmt.str "An error occured trying to fetch %s from albatross: %s"
+                 primary_name err)
+        | Ok (name, unikernel) -> (
+            if unikernel.block_devices <> [] then begin
```
PizieDust: add a TODO/comment about why block devices are not supported yet

unikernel.ml
```diff
@@ -3030,6 +3030,102 @@ struct
...
+                List.map
+                  (fun { Vmm_core.Unikernel.unikernel_device; host_device; _ }
+                     -> (unikernel_device, Some host_device, None))
+                  unikernel.bridges
+              in
+              let argv =
+                match unikernel.argv with
+                | None -> None
+                | Some args ->
+                    Logs.info ~src:Autoscaler.a_logs (fun m ->
```
PizieDust: put this on debug level or remove it entirely

unikernel.ml
```diff
@@ -3030,6 +3030,102 @@ struct
...
+              | Error err -> Lwt.return_error err
+              | Ok cpuid -> (
+                  let config : Vmm_core.Unikernel.config =
+                    {
+                      typ = unikernel.typ;
+                      fail_behaviour = unikernel.fail_behaviour;
+                      startup = unikernel.startup;
+                      memory = unikernel.memory;
+                      block_devices = [];
+                      add_name = false;
```
PizieDust: check properly if `add_name` should indeed be false

unikernel.ml
```diff
@@ -3030,6 +3030,102 @@ struct
...
+                      bridges;
+                      argv;
+                      numcpus = unikernel.numcpus;
```
PizieDust, line 3106: this should be the number of cpuids?

autoscaler.ml
```diff
@@ -156,8 +156,9 @@ module Cluster_manager = struct
             Error (Fmt.str "No primary group found for '%s'." primary_name))
     | None ->
         let g = get_or_create ~user_name clone in
-        g.next_id <- g.next_id + 1;
-        Ok (fst clone)
+        let new_name = Fmt.str "%s-clone-%d" (fst clone) g.next_id in
```
PizieDust: add a TODO to prevent users on mollymawk from naming any unikernel with the suffix `-clone-<ID>`

unikernel.ml
PizieDust: add_name is not returned in `Vmm_core.Unikernel.info` so it's difficult to know if the primary had this as true or false.

autoscaler.ml
```diff
@@ -156,8 +159,9 @@ module Cluster_manager = struct
...
+        add_clone_to_group g (new_name, snd clone) g.next_id;
+        Ok new_name
```
hannesm: so in albatross there are length restrictions for names... now what happens in this code if your `let new_name` produces a name that is longer than allowed by albatross? It looks like `add_clone_to_group` accepts it... but somewhere a Unikernel_create will fail... will this be reflected in your "clone_group"?

unikernel.ml
```diff
@@ -3030,6 +3030,102 @@ struct
...
+    | Error err -> Error err
+    | Ok None -> Error "No policy found for user"
+    | Ok (Some p) ->
+        Ok
+          (let cpu_usage_count = Utils.cpu_usage_count p unikernels in
+           let sorted_cpus =
+             cpu_usage_count
+             |> List.sort (fun (_, c1) (_, c2) -> Int.compare c1 c2)
+           in
+           match sorted_cpus with (id, _) :: _ -> id | [] -> 0)
```
hannesm: 0 may be a CPUid that is not part of the policy, so `[] -> 0` is potentially unsafe (again, leading a `Unikernel_create` command to fail).

unikernel.ml
```diff
@@ -3030,6 +3030,102 @@ struct
...
+              let bridges =
+                List.map
+                  (fun { Vmm_core.Unikernel.unikernel_device; host_device; _ }
+                     -> (unikernel_device, Some host_device, None))
+                  unikernel.bridges
+              in
+              let argv =
+                match unikernel.argv with
+                | None -> None
+                | Some args -> Some (Utils.filter_ip_args args)
```
hannesm: I'm not sure what you try to do. It seems like you strip out some arguments. Why.

What is the general idea of scaling up, and when does it work?
- so in the code there's a "if block_devices <> []"..
- now, once we have a new clone, certainly we would like to distribute load --> this can only happen if we register that e.g. in DNS --> should we assume that the unikernel we try to scale uses DNSvizor (which has the capability to modify DNS)? And if yes, I guess that won't have any "ip" arguments.
- and another thing is, the unikernel name is used by DNSvizor -- so for scaling we somehow need to transport the information that "blog.robur.coop-clone-123" should actually register "blog.robur.coop" in DNS. How can this be done?

unikernel.ml
```diff
@@ -3030,6 +3030,102 @@ struct
...
+              | Error err -> Lwt.return_error err
+              | Ok cpuid -> (
+                  let config : Vmm_core.Unikernel.config =
+                    {
+                      typ = unikernel.typ;
+                      fail_behaviour = unikernel.fail_behaviour;
+                      startup = unikernel.startup;
+                      memory = unikernel.memory;
+                      block_devices = [];
+                      add_name = false;
```
hannesm: I'm not sure about the `add_name` being false.. It's value should be true :) It may be not too well documented, though... And as you mention, maybe not part of Unikernel.info.

unikernel.ml
```diff
@@ -3030,6 +3030,102 @@ struct
...
+                      bridges;
+                      argv;
+                      numcpus = 1;
+                      (* TODO: count the cpuids when they are multiple and not hardcode a value *)
+                      linux_boot_partition = unikernel.linux_boot_partition;
+                      compressed = false;
```
hannesm: Why no compression? Is the image you provide uncompressed?

albatross.ml
```diff
@@ -622,6 +622,43 @@ module Make (S : Tcpip.Stack.V4V6) = struct
         raw_query stack t ~name:vmm_name certificates cmd
           (stats_data vmm_name f)
 
+  let unikernel_get_data name f tls_flow d =
```
hannesm: I'm not sure about this code, has this been tested?

unikernel.ml
```diff
@@ -3030,6 +3030,102 @@ struct
...
+                      let config = { config with compressed } in
+                      let data_stream, push_chunks = Lwt_stream.create () in
+                      let push () = Lwt_stream.get data_stream in
+                      push_chunks (Some binary_string);
+                      push_chunks None;
+                      force_create_unikernel stack albatross
+                        ~unikernel_name:clone_vmm_name ~push config user
+                      >>= function
+                      | Error (`Msg err, _status) -> Lwt.return_error err
+                      | Ok () -> Lwt.return_ok ())
```
hannesm: has this been tested to do chunking (and not loading the entire unikernel image into memory)?

autoscaler.ml
PizieDust: this function no longer produces a string in favor of `Vmm_core.Name.Label.t`

unikernel.ml
PizieDust: > * now, once we have a new clone, certainly we would like to distribute load --> this can only happen if we register that e.g. in DNS --> should we assume that the unikernel we try to scale uses DNSvizor (which has the capability to modify DNS)? And if yes, I guess that won't have any "ip" arguments.

Yes , this is good assumption to make. any unikernels which should be scaled should leave it up to DNSvizor for ip thingies

> * and another thing is, the unikernel name is used by DNSvizor -- so for scaling we somehow need to transport the information that "blog.robur.coop-clone-123" should actually register "blog.robur.coop" in DNS. How can this be done?

I think DNSvizor uses the `--name` arg if we supply one, otherwise it uses the vm name. So in this case, every clone, should have a `--name=primary_name` arg.

unikernel.ml
PizieDust: We use `Unikernel_cmd (Unikernel_get 0)`, I assume 0 means no compression. 
Also, is this the correct way to go? I don't know what is better to do

unikernel.ml
PizieDust: `add_name` is now true

albatross.ml
PizieDust: I haven't done much testing, but will refactor it better

unikernel.ml
PizieDust: I may be wrong, but my understanding is that `Unikernel_get` sends the entire binary as a string in one go. Maybe some update is needed in albatross to support streaming the binary out?

autoscaler.ml
```diff
@@ -156,6 +156,9 @@ module Cluster_manager = struct
     else g
 
   let register_clone group clone =
+    (* TODO: add a validation rule during unikernel creation so users 
+            can't name any vm with the suffix "-clone-INT". 
+            What about unikernels deployed directly on albatross? *)
```
hannesm, line 161: is this something to be addressed in this PR, or later? certainly "what about unikernels deployed directly on albatross" -- you can't do anything in mollymawk about that.

unikernel.ml
```diff
@@ -3030,6 +3045,87 @@ struct
...
+                  unikernel.bridges
+              in
+              let argv =
+                let name_arg =
+                  "--name=" ^ Configuration.name_to_str unikernel_name
+                in
+                match unikernel.argv with
+                | None -> Some [ name_arg ]
+                | Some args -> Some (name_arg :: args)
+              in
```
hannesm: this potentially adds a second `--name` argument?

Also, thinking about the metrics reporting, we will then have something known to albatross as `hello-clone-1`, but reporting as `hello` itself -- I think we need to think a bit harder about this. Maybe such a unikernel needs several names -- the "hostname" (under which it operates) and the "monitoring name" (how it reports statistics).

What do you think?

unikernel.ml
hannesm: It is a tradeoff between CPU and network bandwidth. I'd recommend to use `Unikernel_get 4` (or actually, try different values between 0 and 9 -- see how it compresses and how much CPU is used for it).

Since we're on a remote end of a connection, I'd use compression and at least 4 to save bandwidth.

unikernel.ml
```diff
@@ -3030,6 +3045,87 @@ struct
...
+                      memory = unikernel.memory;
+                      block_devices = [];
+                      add_name = true;
+                      bridges;
+                      (* we assume for unikernels which are scaled, their ip is configured via DNSvizor *)
+                      argv;
+                      numcpus = unikernel.numcpus;
+                      linux_boot_partition = unikernel.linux_boot_partition;
+                      compressed = false;
+                      image = "";
```
hannesm, line 3109: now, since you already have the image at hand (see below), you can as well provide it here directly.

It indeed seems a unikernel_get will not stream, but hand you back the entire unikernel image. I think this is fine for now (and let's use that in mollymawk instead of trying to do streaming).

unikernel.ml
hannesm: yes, but for the time being, maybe just use the image you got. Going down the rabbit hole to update albatross will take some more time.

albatross.ml
PizieDust: will just get rid of this and use `albatross_state.query`

autoscaler.ml
PizieDust: this will be addressed later.

unikernel.ml
PizieDust: yes. to DNSvizor, both hello and hello-clone-1 needs to be registered with the same hostname, but to everything else, like for grafana,  it's better they have separate names

albatross.ml
hannesm: uhm, now? or after this PR is merged?

unikernel.ml
hannesm: so what will be done?

unikernel.ml
```diff
@@ -3030,6 +3045,108 @@ struct
...
+                          Lwt.return_ok () >>= function
+                          | Error err ->
+                              Log.err (fun m ->
+                                  m
+                                    "spawn_clone: Error getting binary for %s: \
+                                     %s"
+                                    (Configuration.name_to_str unikernel_name)
+                                    err);
+                              Lwt.return_error err
+                          | Ok () -> Lwt.return_ok ()))
```
hannesm: I don't quite understand this block of code? -- isn't above already the code path for "error getting binary"? And here a "Lwt.return_ok () >>= function | Error _ -> ... | Ok () -> ..." -- i.e. where the Error case in the function is never being executed?

albatross.ml
hannesm: ah, this has been addressed.

unikernel.ml
PizieDust: so I was thinking we could add a --metrics-name but this means recompiling all vms to use it.....

another option, I don't know if DNSvizor supports it but `Dynamic Updates in the Domain Name System (DNS UPDATE)` which is rfc 2136, but may mean mollymawk will have to be able to:
- deploy the vm and let albatross add the --name thingy
- wait for the vm to boot or perhaps an albatross log emitted to trigger mollymawk to request what ip DNSvizor gave it
- then mollymawk sends back a dns update request to DNSvizor

what do you think about this approach?

PizieDust: and our `ocaml-dns` already supports rfc 2136 https://github.com/mirage/ocaml-dns#supported-rfcs

PizieDust:
@hannesm while testing this, and using just `albatross_state.query`, I get: 

```
2026-07-03T15:14:10-00:00: [INFO] [autoscaling] robur/blog: overloaded (100%), spawning new clone: blog-clone-1
2026-07-03T15:14:10-00:00: [ERROR] [autoscaling] Error getting binary for blog: albatross short reply: received 16384 bytes, expected 4659061 bytes
2026-07-03T15:14:10-00:00: [ERROR] [autoscaling] Error handling stats for robur:blog: albatross short reply: received 16384 bytes, expected 4659061 bytes
2026-07-03T15:14:20-00:00: [INFO] [autoscaling] robur/blog: cooldown (usage: 0%)

```

but i just saw another bug, it enters cooldown even when the new clone was never deployed. this should be easy to fix 

but i think the push function we have is just for uploading data to albatross. need to figure out a way to get the entire response from albatross

PizieDust:
> @hannesm while testing this, and using just `albatross_state.query`, I get:
> 
> ```
> 2026-07-03T15:14:10-00:00: [INFO] [autoscaling] robur/blog: overloaded (100%), spawning new clone: blog-clone-1
> 2026-07-03T15:14:10-00:00: [ERROR] [autoscaling] Error getting binary for blog: albatross short reply: received 16384 bytes, expected 4659061 bytes
> 2026-07-03T15:14:10-00:00: [ERROR] [autoscaling] Error handling stats for robur:blog: albatross short reply: received 16384 bytes, expected 4659061 bytes
> 2026-07-03T15:14:20-00:00: [INFO] [autoscaling] robur/blog: cooldown (usage: 0%)
> ```
> 
> but i just saw another bug, it enters cooldown even when the new clone was never deployed. this should be easy to fix
> 
> but i think the push function we have is just for uploading data to albatross. need to figure out a way to get the entire response from albatross

So I am testing this with a tool called [wrk](https://github.com/wg/wrk) which is an http benchmarking tool. i just use it to dump a bunch of traffic on the ip of the vm and then see if mollymawk reacts as expected to the cpu spikes. 

now it doesnt enter a cooldown if deploying a new vm fails. 
```
2026-07-03T15:44:00-00:00: [INFO] [autoscaling] robur/blog: high usage (100%), [ticks: 1/3]
2026-07-03T15:44:10-00:00: [INFO] [autoscaling] robur/blog: high usage (100%), [ticks: 2/3]
2026-07-03T15:44:20-00:00: [INFO] [autoscaling] robur/blog: overloaded (100%), spawning new clone: blog-clone-1
2026-07-03T15:44:21-00:00: [ERROR] [autoscaling] Error getting binary for blog: albatross short reply: received 16384 bytes, expected 4659061 bytes
2026-07-03T15:44:21-00:00: [ERROR] [autoscaling] Error handling stats for robur:blog: albatross short reply: received 16384 bytes, expected 4659061 bytes
2026-07-03T15:44:30-00:00: [INFO] [autoscaling] robur/blog: high usage (100%), [ticks: 1/3]
2026-07-03T15:44:40-00:00: [INFO] [autoscaling] robur/hello: cooldown (usage: 0%)
2026-07-03T15:44:40-00:00: [INFO] [autoscaling] robur/blog: high usage (100%), [ticks: 2/3]
^Csolo5-hvt: Exiting on signal 2
```

PizieDust:
we have this issue:

```
2026-07-03T16:58:40-00:00: [WARNING] [application] albatross buffer too small: 16380 bytes, requires 4659057 bytes
Fatal error: exception Invalid_argument("String.sub / Bytes.sub")
Raised at Stdlib.invalid_arg in file "stdlib.ml", line 30, characters 20-45
Called from Stdlib__String.sub in file "string.ml", line 44, characters 2-23
Called from Dune__exe__Albatross.Make.split_many.split in file "albatross.ml", line 260, characters 17-46
Called from Dune__exe__Albatross.Make.split_many in file "albatross.ml", line 274, characters 4-19
Called from Dune__exe__Albatross.Make.continue_reading in file "albatross.ml", line 278, characters 15-27
Called from Lwt.Sequential_composition.bind.create_result_promise_and_callback_if_deferred.callback in file "duniverse/lwt/src/core/lwt.ml", line 1853, characters 16-19
Solo5: solo5_exit(2) called
```

this happens when we try to get the binary from albatross, somewhere in `continue_reading` and `split_many`, it fails. I think these two functions were designed for self-contained packets in a stream

unikernel.ml
hannesm: I don't understand the approach you have in mind. DNSvizor is able to issue DNS update packets... now you want to move that functionality to mollymawk? how would mollymawk get the IP address of the unikernel?

PizieDust: oh nice, i didn't know DNSvizor could do that, 
for the IP, the idea was to use `dns_client_mirage.gethostbyname` and see what IP dnsvizor assigned to it, we already currently use this for configuring monitoring and metrics.  so the same approach can be used to get an ip of a unikernel when DNSvizor has assigned an IP. 

I will prefer to not have the DNS update packet functionality in mollymawk, but I don't know how to make DNSvizor do this

PizieDust:
> we have this issue:
> 
> ```
> 2026-07-03T16:58:40-00:00: [WARNING] [application] albatross buffer too small: 16380 bytes, requires 4659057 bytes
> Fatal error: exception Invalid_argument("String.sub / Bytes.sub")
> Raised at Stdlib.invalid_arg in file "stdlib.ml", line 30, characters 20-45
> Called from Stdlib__String.sub in file "string.ml", line 44, characters 2-23
> Called from Dune__exe__Albatross.Make.split_many.split in file "albatross.ml", line 260, characters 17-46
> Called from Dune__exe__Albatross.Make.split_many in file "albatross.ml", line 274, characters 4-19
> Called from Dune__exe__Albatross.Make.continue_reading in file "albatross.ml", line 278, characters 15-27
> Called from Lwt.Sequential_composition.bind.create_result_promise_and_callback_if_deferred.callback in file "duniverse/lwt/src/core/lwt.ml", line 1853, characters 16-19
> Solo5: solo5_exit(2) called
> ```
> 
> this happens when we try to get the binary from albatross, somewhere in `continue_reading` and `split_many`, it fails. I think these two functions were designed for self-contained packets in a stream

this works now and clones correctly:

```
64
2026-07-03T19:03:22-00:00: [INFO] [application] Domain from DHCP lease: management
2026-07-03T19:03:22-00:00: [INFO] [application] Initialise an HTTP server (no HTTPS) on port 8080
2026-07-03T19:03:22-00:00: [INFO] [application] Next background update in 17797 seconds
2026-07-03T19:03:50-00:00: [INFO] [autoscaling] robur/blog: high usage (100%), [ticks: 1/3]
2026-07-03T19:04:00-00:00: [INFO] [autoscaling] robur/blog: high usage (100%), [ticks: 2/3]
2026-07-03T19:04:10-00:00: [INFO] [autoscaling] robur/blog: overloaded (100%), spawning new clone: blog-clone-1
2026-07-03T19:04:10-00:00: [INFO] [autoscaling] config is: typ solo5
                                                compression true image 0 bytes
                                                fail behaviour restart all except 1
                                                startup at not specified
                                                add_name true CPUids 2
                                                128 MB memory block devices 
                                                bridge service -> service, management -> management
                                                #cpus 1
                                                linux_boot not specified
2026-07-03T19:04:11-00:00: [INFO] [application] albatross-force-create: successfully created blog-clone-1 with result \"created unikernel\"
2026-07-03T19:04:11-00:00: [INFO] [autoscaling] Successfully cloned blog to blog-clone-1
```

unikernel.ml
PizieDust: I just noticed, when a unikernel has the `--name` argument, it also get's copied to the clone. so we may need a function to remove it before constructing the config for the clone.

PizieDust: @hannesm  so, this is the situation
- clones must have unique names (blog-clone-1) so metrics and syslog don't become a tangled mess of colliding data
- if a Unikernel has multiple --name arguments, it crashes with Exit 64 and then the `--name` one is used, but we acutally want to strip this argument out when creating clones. 

so to solve the issue of having unique clone names, we can do a tiny update in DNSvizor so it recognizes mollymawk clone names. Does that  make sense?

```ocaml
(* In DNSvizor's dhcp_lease process, right when it receives the Hostname *)
let base_name = 
  match String.split_on_char '-' name with
  | base :: "clone" :: _id -> base 
  | _ -> name
in
(* Now insert the IP into `base_name.domain` instead of `name.domain` *)

```

if we do this update, then i think it solves the problem. but only if we want this update in DNSvizor

PizieDust: here is the section in DNSvizor. https://github.com/robur-coop/dnsvizor/blob/main/unikernel.ml#L2117-L2123

~i also just saw that DNSvizor can update tlstunel. this is super cool~ , seems it overrides

PizieDust: I opened https://github.com/robur-coop/dnsvizor/pull/137 on DNSvizor to recognize mollymawk clones

unikernel.ml
```diff
@@ -3079,6 +3079,16 @@ struct
...
               | Ok cpuid -> (
+                  let argv =
+                    Option.fold unikernel.argv ~none:None ~some:(fun arg ->
+                        Some
+                          (List.fold_left
+                             (fun acc arg ->
+                               if String.starts_with ~prefix:"--name=" arg then
+                                 acc
+                               else arg :: acc)
+                             [] arg))
```
hannesm: you could use Option.map instead.

I don't think this works as intended... so, back at the argument vector: this is a list of strings. They're all concatenated and passed to the unikernel. arguments can be passed as _a single string_ `[ "--ipv4=1.2.3.4 --ipv4-gateway=1.2.3.1" ]` or two string `[ "--ipv4=1.2.3.4" ; "--ipv4-gateway=1.2.3.1" ]`... that's at least what I remember, maybe worth to check...

hannesm:
Thanks, so Reynir and myself decided that the DNSvizor change is fine for now. This PR is then ready to be merged as well as far as I can see -- apart from the "filter of --name" part I commented above. Would you be able to fix that part?

unikernel.ml
PizieDust: thanks for catching that

PR #257, opened 2026-06-18
Restart unikernels with modified arguments

### LINKED ISSUE #35: Updating unikernels
Provide UI and functionality to enable updating information about unikernels such as:
- changing/updating arguments
- updating block devices

PizieDust:
We can restart a unikernel and modify it's json arguments. 
In the future, we can have a UI builder which takes such a json and renders UI elements similar to what the deploy form looks like (see https://github.com/robur-coop/mollymawk/issues/258)

Currently, the memory value gets ignored during the restart: see https://github.com/robur-coop/albatross/issues/275

Closes https://github.com/robur-coop/mollymawk/issues/35

### DIFF
--- a/albatross_json.ml
+++ b/albatross_json.ml
@@ -409,3 +409,39 @@ let config_to_json (cfg : Vmm_core.Unikernel.config) =
           ~some:(fun s -> `String s)
           cfg.linux_boot_partition );
     ]
+
+let unikernel_info_to_arguments_json info =
+  `Assoc
+    [
+      ("typ", `String (typ info.Vmm_core.Unikernel.typ));
+      ("fail_behaviour", fail_behaviour info.fail_behaviour);
+      ("startup", `Int (Option.fold ~none:50 ~some:Fun.id info.startup));
+      (* add_name is not found in Vmm_core.Unikernel.info *)
+      ("add_name", `Bool false);
+      ("cpuids", cpuids info.cpuids);
+      ("memory", memory info.memory);
+      ("block_devices", block_devices info.block_devices);
+      ("bridges", bridges info.bridges);
+      ("arguments", argv info.argv);
+      ("numcpus", `Int info.numcpus);
+      ( "linux_boot_partition",
+        `String (Option.fold ~none:"" ~some:Fun.id info.linux_boot_partition) );
+    ]
+
+let arguments_of_json str =
+  match config_of_json str with
+  | Ok config ->
+      Ok
+        {
+          Vmm_core.Unikernel.fail_behaviour = config.fail_behaviour;
+          add_name = config.add_name;
+          startup = config.startup;
+          cpuids = config.cpuids;
+          memory = config.memory;
+          block_devices = config.block_devices;
+          bridges = config.bridges;
+          argv = config.argv;
+          numcpus = config.numcpus;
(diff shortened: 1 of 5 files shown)


albatross_json.ml
```diff
@@ -358,6 +357,8 @@ let config_of_json str =
       linux_boot_partition;
     }
 
+let typ typ = match typ with `Solo5 -> "solo5" | `BHyve -> "bhyve"
```
hannesm: could be `let typ = function ...`

albatross_json.ml
```diff
@@ -409,3 +409,39 @@ let config_to_json (cfg : Vmm_core.Unikernel.config) =
...
     ]
+
+let unikernel_info_to_arguments_json info =
+  `Assoc
+    [
+      ("typ", `String (typ info.Vmm_core.Unikernel.typ));
+      ("fail_behaviour", fail_behaviour info.fail_behaviour);
+      ("startup", `Int (Option.fold ~none:50 ~some:Fun.id info.startup));
+      (* add_name is not found in Vmm_core.Unikernel.info *)
+      ("add_name", `Bool false);
```
hannesm: add_name should be true.

unikernel.ml
```diff
@@ -2053,59 +2053,35 @@ struct
...
+                  ~data:(`String ("Error querying albatross: " ^ msg))
+                  `Internal_server_error
+            | Ok (_hdr, res) -> (
+                Albatross.set_online albatross;
+                match Albatross_json.res res with
+                | Ok res -> Middleware.http_response reqd ~data:res `OK
+                | Error (`String err) ->
+                    Middleware.http_response reqd ~data:(`String err)
+                      `Internal_server_error)))
     | _ ->
```
hannesm: could be directly `None`.

hannesm:
I don't quite understand the code, but is it the case that it is still possible to restart without modifying the arguments? And if yes, isn't line 2082 dealing with "there have been no arguments transmitted"? -- which at the moment leads to a failure.

Or, how is the flow of data with no changed arguments?

PizieDust:
> I don't quite understand the code, but is it the case that it is still possible to restart without modifying the arguments? And if yes, isn't line 2082 dealing with "there have been no arguments transmitted"? -- which at the moment leads to a failure.
> 
> Or, how is the flow of data with no changed arguments?

Yes we can restart with no changed arguments. 
So on the UI we have a switch toggle which displays a text area for us to modify the arguments. 

If the switch is off, then the original arguments are transmitted.

hannesm:
> If the switch is off, then the original arguments are transmitted.

And, would it be possible to instead not transmit arguments? That way I think albatross could do less work (no need to dump the state file)?

PizieDust:
> > If the switch is off, then the original arguments are transmitted.
> 
> And, would it be possible to instead not transmit arguments? That way I think albatross could do less work (no need to dump the state file)?

Yes I think this is possible

PizieDust:
> > If the switch is off, then the original arguments are transmitted.
> 
> And, would it be possible to instead not transmit arguments? That way I think albatross could do less work (no need to dump the state file)?

This is done now :) @hannesm

PizieDust:
Changed the request from the default `multipart/form` enctype to `application/json` so the API system can still work.

PR #262, opened 2026-07-20
Read full tls packets

PizieDust:
If the data is larger than a TLS packet, we should loop till we read all the data.

cc @reynir @hannesm


hannesm:
I didn't quite know how to put in suggestions, but I reworked this to be less code, please take a look at the last commit of #264

PizieDust:
I will close this in favor of https://github.com/robur-coop/mollymawk/pull/264

PR #264, opened 2026-07-20
Make raw query smaller

hannesm:
on top of #262 //cc @PizieDust


PizieDust (approved):
Thanks @hannesm

reynir:
I pushed a commit so we use a buffer for accumulating the data instead of repeated string concatenation (quadratic).

hannesm:
nice @reynir

hannesm:
I guess this is ready to go, no?

reynir (approved):

PR #267, opened 2026-07-27
Restart with arguments

PizieDust:
Wen we restart with arguments, `network_interfaces` are under the key `bridges`.

### DIFF
--- a/albatross_json.ml
+++ b/albatross_json.ml
@@ -279,7 +279,7 @@ let config_of_json str =
       (get "memory" dict)
   in
   let* bridges =
-    let brdige bs =
+    let bridge bs =
       List.fold_left
         (fun r x ->
           let* r = r in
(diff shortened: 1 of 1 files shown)


hannesm:
I don't understand this PR. Can you please explain a bit more context?

What does "under the key 'bridges'" mean?

albatross_json.ml
```diff
@@ -279,15 +279,21 @@ let config_of_json str =
       (get "memory" dict)
   in
   let* bridges =
+    let brdige bs =
```
hannesm: there's a typo "brdige".

hannesm:
Previously, there was potentially a "network_devices" key in the json present. Now, if this is not present, another key, "bridges" is looked up.

I wonder who generates the json and why should we support both keys?

PizieDust:
> Previously, there was potentially a "network_devices" key in the json present. Now, if this is not present, another key, "bridges" is looked up.
> 
> I wonder who generates the json and why should we support both keys?

Thanks you're right. It was just a simple change in the ` unikernel_info_to_arguments_json` which needed the change.

PR #273, opened 2026-08-21
Deprecate v1 to v8 of storage data format

PizieDust:
This PR deprecates v1 to v8 of the data format stored on the block device. Related code has been removed as well.


storage.ml
```diff
@@ -37,51 +30,33 @@ let t_of_json json =
...
                 let* user =
-                  if v = 1 then User_model.user_v1_of_json js
-                  else if v = 2 || v = 3 then User_model.user_v2_of_json js
-                  else if v = 4 || v = 5 then
-                    User_model.(user_v3_of_json cookie_v1_of_json) js
-                  else if v = 6 then
-                    User_model.(user_v4_of_json cookie_v1_of_json) js
-                  else if v = 7 || v = 8 || v = 9 then
-                    User_model.(user_v5_of_json cookie_of_json) js
+                  if v = 9 then User_model.(user_old_of_json cookie_of_json) js
```
hannesm: since there may be at some point newer versions, this `old_of_json` will be a strange name. Why not keep the version information in there? I.e. `v9_of_json`?

PR #274, opened 2026-08-21
Move libriaries to src

PizieDust:
This Pr moves some files into `src` directory for better organization and to lay the groundwork for writing tests.


src/dune
```diff
@@ -0,0 +1,4 @@
+(library
+ (name mollymawk_libraries)
+ (public_name mollymawk_libraries)
```
reynir: Not sure we need a public_name (but it's not a big deal)

dune-project
```diff
@@ -0,0 +1,2 @@
+(lang dune 2.9)
+(name mollymawk_libraries)
```
reynir, line 2: Does mirage overwrite this?

reynir (commented):
Would you like me to test this? Looks fine to me

dune-project
PizieDust: I did a `mirage-clean` and a `mirage confiure -t hvt` and it didn't overwrite it. :)

PizieDust:
Thanks @hannesm and @reynir

PR #275, opened 2026-08-21
Add tests for serializing data stored to the block device

PizieDust:
In this PR, we move the functor code for connecting to the block device and reading/writing data into unikernel.ml and move the remaining storage functions in src.
We add some tests to check various `from_json` and `to_json` code. 

This is a first phase. More refactoring and tests will be added subsequently.

Test results: 

```sh
$ dune runtest
Testing `Mollymawk data serialization tests for storage'.
This run has ID `GSL0YUZ0'.

  [OK]          Version tests                   0   Deprecated version.
  [OK]          Version tests                   1   Accepted version.
  [OK]          Version tests                   2   Invalid version.
  [OK]          Email config tests              0   Email configuration in v9.
  [OK]          Email config tests              1   No email configuration in v9.
  [OK]          Email config tests              2   Email configuration in v10.
  [OK]          Email config tests              3   No email configuration in v10.
  [OK]          Albatross config tests          0   No albatross configuration in v9.
  [OK]          Albatross config tests          1   No albatross configuration in v10.
  [OK]          Albatross config tests          2   Valid albatross configuration in v9.
  [OK]          Albatross config tests          3   Valid albatross configuration in v10.
  [OK]          Albatross config tests          4   Incompatible private key and certificate combination.
  [OK]          Albatross config tests          5   Empty certificate.
  [OK]          Albatross config tests          6   Empty private_key.

Full test results in `~/...'
Test Successful in 0.001s. 14 tests run.
```


unikernel.ml
```diff
@@ -32,6 +32,268 @@ struct
...
+            user.User_model.email_verification_uuid)
+        store.users
+
+    let count_active store =
+      List.length (List.filter (fun u -> u.User_model.active) store.users)
+
+    let count_superusers store =
+      List.length (List.filter (fun u -> u.User_model.super_user) store.users)
+  end
+
```
reynir: To keep the diff and unikernel.ml can't you keep this in storage.ml? I mean not the storage.ml in `src/` but one at the root. Then in the storage.ml at the root you can `include Storage` and the interface should be the same I presume.

tests/storage/tests.ml
```diff
@@ -0,0 +1,331 @@
+Mirage_crypto_rng_unix.use_default ();;
+
+let of_string_exn f =
+  match f with Ok a -> a | Error (`Msg err) -> failwith err
+
+let label_of_string_exn label =
+  of_string_exn (Vmm_core.Name.Label.of_string label)
+
+let ip_of_string_exn ip = of_string_exn (Ipaddr.of_string ip)
```
reynir: There is already `Ipaddr.of_string_exn`

tests/storage/tests.ml
```diff
@@ -0,0 +1,331 @@
...
+let email_of_string_exn email = of_string_exn (Mrmime.Mailbox.of_string email)
+
+let signing_request_exn pk =
+  match X509.Signing_request.create [] pk with
+  | Ok c -> c
+  | Error _ -> failwith "invalid signing request"
+
+let private_key = X509.Private_key.generate `ED25519
+let second_private_key = X509.Private_key.generate `ED25519
```
reynir: Consider using a seed (`?seed` argument) so we generate the same key every time. Unless you have reasons not to. This will help making the tests less random.

tests/storage/tests.ml
```diff
@@ -0,0 +1,331 @@
...
+      && String.equal
+           (X509.Public_key.fingerprint
+              (X509.Certificate.public_key c1.certificate))
+           (X509.Public_key.fingerprint
+              (X509.Certificate.public_key c2.certificate))
+      && String.equal
+           (X509.Public_key.fingerprint
+              (X509.Private_key.public c1.private_key))
+           (X509.Public_key.fingerprint
+              (X509.Private_key.public c2.private_key))
```
reynir: Could be refactored. For example, they share comparison of public keys.

tests/storage/tests.ml
```diff
@@ -0,0 +1,331 @@
...
+  | c1 :: _, c2 :: _ ->
+      Vmm_core.Name.Label.equal c1.name c2.name
+      && Ipaddr.compare c1.server_ip c2.server_ip = 0
+      && c1.server_port = c2.server_port
+      && Ptime.equal c1.updated_at c2.updated_at
+      && String.equal
+           (X509.Public_key.fingerprint
+              (X509.Certificate.public_key c1.certificate))
+           (X509.Public_key.fingerprint
+              (X509.Certificate.public_key c2.certificate))
```
reynir: Maybe comparing just the public key of the certificate is not sufficient? You can get a fingerprint of the certificate and compare those.

tests/storage/tests.ml
```diff
@@ -0,0 +1,331 @@
...
+           (X509.Public_key.fingerprint
+              (X509.Private_key.public c2.private_key))
+  | _ -> false
+
+let cmp_option f o1 o2 =
+  match (o1, o2) with
+  | None, None -> true
+  | None, _ -> false
+  | Some _, None -> false
+  | Some a, Some b -> f a b
```
reynir: This is `Option.equal`

tests/storage/tests.ml
```diff
@@ -0,0 +1,331 @@
...
+      && Mrmime.Mailbox.equal u1.email u2.email
+      && String.equal u1.password u2.password
+      && String.equal u1.uuid u2.uuid
+      && u1.active = u2.active
+      && u1.super_user = u2.super_user
+      && Ptime.equal u1.updated_at u2.updated_at
+      && Ptime.equal u1.created_at u2.created_at
+      && cmp_option Ptime.equal u1.email_verified u2.email_verified
+      && cmp_option Uuidm.equal u1.email_verification_uuid
+           u2.email_verification_uuid
```
reynir: This is a bit brittle to if we add new fields to `User_model.user`. Consider destructuring them

unikernel.ml
PizieDust: yes we can do this. more functions here will be removed, especially those which don't connect, write or read the block device and these functions will be moved to src/storage.ml

PizieDust: this is now done :)

tests/storage/tests.ml
PizieDust: this is now done :)

tests/storage/tests.ml
PizieDust: we get the fingerprints of both certificates and private_keys

tests/storage/tests.ml
PizieDust: thanks

tests/storage/tests.ml
```diff
@@ -0,0 +1,355 @@
...
+let signing_request_exn pk =
+  match X509.Signing_request.create [] pk with
+  | Ok c -> c
+  | Error _ -> failwith "invalid signing request"
+
+let private_key =
+  X509.Private_key.generate ~seed:"robur_in_essaouira_2026" `ED25519
+
+let second_private_key =
+  X509.Private_key.generate ~seed:"robur_in_essaouira_20_26" `ED25519
```
hannesm: the first and second key is the same now, is this intentional? they're using the same seed.

tests/storage/tests.ml
```diff
@@ -0,0 +1,355 @@
...
+  Fmt.pf ppf "%a" Yojson.Basic.pp (Storage.t_to_json users configuration email)
+
+let eq_key k1 k2 =
+  String.equal (X509.Public_key.fingerprint k1) (X509.Public_key.fingerprint k2)
+
+let eq_config (config_1 : Configuration.t list)
+    (config_2 : Configuration.t list) =
+  match (config_1, config_2) with
+  | [], [] -> true
+  | c1 :: _, c2 :: _ ->
```
hannesm: didn't @reynir mention, you could destruct here the configs, ` { name ; server_port ; ... }, { name = namea; server_port = server_porta ; ... }` -- so that once there are more fields in configuration, it will be noticed (compilation warning)

tests/storage/tests.ml
```diff
@@ -0,0 +1,355 @@
...
+      Vmm_core.Name.Label.equal c1.name c2.name
+      && Ipaddr.compare c1.server_ip c2.server_ip = 0
+      && c1.server_port = c2.server_port
+      && Ptime.equal c1.updated_at c2.updated_at
+      && eq_key
+           (X509.Certificate.public_key c1.certificate)
+           (X509.Certificate.public_key c2.certificate)
```
hannesm: the certificate has more than a public key, so I'd use `String.equal (X509.Certificate.fingerprint `SHA256 c1.certificate) (X509.Certificate.fingerprint `SHA256 c2.certificate)`

tests/storage/tests.ml
```diff
@@ -0,0 +1,355 @@
...
+      && Option.equal Ptime.equal ev1 ev2
+      && Option.equal Uuidm.equal evu1 evu2
+  | _ -> false
+
+let eq_emails (e1 : Utils.Email.t) (e2 : Utils.Email.t) =
+  Ipaddr.compare e1.server e2.server = 0
+  && e1.port = e2.port
+  && String.equal e1.base_url e2.base_url
+  && Mrmime.Mailbox.equal e1.from_email e2.from_email
+  && Option.equal Mrmime.Mailbox.equal e1.to_email e2.to_email
```
hannesm: same as above (as you did for users): destruct...

tests/storage/tests.ml
```diff
@@ -0,0 +1,355 @@
...
+
+let eq_storage (u1, c1, e1) (u2, c2, e2) =
+  eq_users u1 u2 && eq_config c1 c2 && Option.equal eq_emails e1 e2
+
+let storage_t = Alcotest.testable pp_storage eq_storage
+
+let mock_storage ?(version = 10) ?(users = []) ?(configuration = [])
+    ?(email = None) () =
+  Storage.t_to_json ~version users configuration email
```
hannesm, line 147: with this approach, what you check is "some high level value", which is then converted to json, and back to the high level value.

but what is not checked is the actual bytes on disk. I'd appreciate to have some testing for that -- so maybe in an interactive session construct the bytes/string that results from t_to_json, and have this as a strung in the testsuite.

hannesm: as well as invalid json lead to errors

tests/storage/tests.ml
```diff
@@ -0,0 +1,355 @@
...
+let check_invalid_version () =
+  let expected =
+    `Msg
+      "expected version 10, found version 11. note: version [1 - 8] is now \
+       deprecated."
+  in
+  Alcotest.(
+    check (result storage_t msg_t)
+      "mollymawk should fail to start when version is invalid" (Error expected)
+      (Storage.t_of_json (mock_storage ~version:11 ())))
```
hannesm: that's brittle since once a new version is inroduced, it will start to succeed. I'd either remove or use a very high version (1000?) which we'll unlkely reach

tests/storage/tests.ml
```diff
@@ -0,0 +1,355 @@
...
+      check_valid_albatross_config_in_v9 );
+    ( "Valid albatross configuration in v10",
+      `Quick,
+      check_valid_albatross_config_in_v10 );
+    ( "Incompatible private key and certificate combination",
+      `Quick,
+      check_invalid_private_key_in_albatross_config );
+    ("Empty certificate", `Quick, check_missing_certificate_in_albatross_config);
+    ("Empty private_key", `Quick, check_missing_private_key_in_albatross_config);
+  ]
```
hannesm, line 423: maybe multiple albatross configurations are worth testing as well (some which are good, some where the name is equal (which I guess should be rejected))

store.ml
```diff
@@ -0,0 +1,116 @@
...
+
+module Make (BLOCK : Mirage_block.S) = struct
+  module Stored_data = OneFFS.Make (BLOCK)
+  open Lwt.Infix
+
+  type t = {
+    disk : Stored_data.t;
+    mutable users : User_model.user list;
+    mutable configurations : Configuration.t list;
+    mutable email : Utils.Email.t option;
```
hannesm: I'd move this record to `storage.ml` -- without the disk -- and not define any type in this module. the `connect` below then returns a pair `Stored_data.t * Storage.t`.

store.ml
```diff
@@ -0,0 +1,116 @@
+include Storage
```
hannesm: why is that? can we remove the include and just use `Storage` explicitly?

store.ml
```diff
@@ -0,0 +1,116 @@
...
+        if not exists then
+          Lwt.return
+            (Storage.error_msgf "configuration %s not found"
+               (Configuration.name_to_str configuration.name))
+        else
+          let configurations =
+            List.map
+              (fun c -> if name_eq c then configuration else c)
+              t.configurations
+          in
```
hannesm: as I mentioned yesterday, the marked piece can go into Storage... and can then be tested (but maybe for later?)

tests/storage/tests.ml
PizieDust: there is a slight difference in the way the year is written. one is 2026 and the other is 20_26

hannesm: ah, that was not very obvious to me... I'd use an entirely different string

tests/storage/tests.ml
PizieDust: we can do more of the tests in a separate PR so as to keep this one from getting too large :)

tests/storage/tests.ml
```diff
@@ -0,0 +1,435 @@
...
+
+let msg_t =
+  let pp ppf (`Msg s) = Fmt.string ppf s in
+  Alcotest.testable pp (fun (`Msg a) (`Msg b) -> String.equal a b)
+
+let pp_storage ppf (users, configuration, email) =
+  Fmt.pf ppf "%a" Yojson.Basic.pp (Storage.t_to_json users configuration email)
+
+let eq_key k1 k2 = String.equal k1 k2
+
```
hannesm:
```suggestion
```

tests/storage/tests.ml
```diff
@@ -0,0 +1,435 @@
...
+        updated_at = ua2;
+        certificate = cer2;
+        private_key = pk2;
+      } =
+        c2
+      in
+      Vmm_core.Name.Label.equal n1 n2
+      && Ipaddr.compare si1 si2 = 0
+      && sp1 = sp2 && Ptime.equal ua1 ua2
+      && eq_key
```
hannesm:
```suggestion
      && String.equal
```

tests/storage/tests.ml
```diff
@@ -0,0 +1,435 @@
...
+           (X509.Certificate.fingerprint `SHA256 cer1)
+           (X509.Certificate.fingerprint `SHA256 cer2)
+      && eq_key
```
hannesm:
```suggestion
      && String.equal
```

tests/storage/tests.ml
```diff
@@ -0,0 +1,435 @@
...
+        server_port = sp2;
+        updated_at = ua2;
+        certificate = cer2;
+        private_key = pk2;
+      } =
+        c2
+      in
+      Vmm_core.Name.Label.equal n1 n2
+      && Ipaddr.compare si1 si2 = 0
+      && sp1 = sp2 && Ptime.equal ua1 ua2
```
hannesm:
```suggestion
      && Int.equal sp1 sp2 && Ptime.equal ua1 ua2
```

tests/storage/tests.ml
```diff
@@ -0,0 +1,435 @@
...
+        tokens = _tk2;
+        cookies = _ck2;
+        unikernel_updates = _uk2;
+        scaling_policies = _sp2;
+      } =
+        u2
+      in
+      Vmm_core.Name.Label.equal n1 n2
+      && Mrmime.Mailbox.equal e1 e2 && String.equal p1 p2
+      && String.equal id1 id2 && a1 = a2 && s1 = s2 && Ptime.equal uat1 uat2
```
hannesm:
```suggestion
      && String.equal id1 id2 && Bool.equal a1 a2 && Bool.equal s1 s2 && Ptime.equal uat1 uat2
```

tests/storage/tests.ml
```diff
@@ -0,0 +1,435 @@
...
+    Utils.Email.server = s2;
+    port = p2;
+    base_url = bu2;
+    from_email = fe2;
+    to_email = te2;
+  } =
+    e2
+  in
+  Ipaddr.compare s1 s2 = 0
+  && p1 = p2 && String.equal bu1 bu2
```
hannesm:
```suggestion
  && Int.equal p1 p2 && String.equal bu1 bu2
```

hannesm:
Looks fine to me, just some very small things to avoid polymorphic equality (`=`) in the tests. :)

tests/storage/tests.ml
hannesm: sure, sounds good to me.

tests/storage/tests.ml
```diff
@@ -0,0 +1,435 @@
...
+      } =
+        u2
+      in
+      Vmm_core.Name.Label.equal n1 n2
+      && Mrmime.Mailbox.equal e1 e2 && String.equal p1 p2
+      && String.equal id1 id2 && a1 = a2 && s1 = s2 && Ptime.equal uat1 uat2
+      && Ptime.equal cat1 cat2
+      && Option.equal Ptime.equal ev1 ev2
+      && Option.equal Uuidm.equal evu1 evu2
+  | _ -> false
```
hannesm: in the future (when testing multiple users), this likely need to be adjusted...

tests/storage/tests.ml
```diff
@@ -0,0 +1,435 @@
...
+      Vmm_core.Name.Label.equal n1 n2
+      && Ipaddr.compare si1 si2 = 0
+      && sp1 = sp2 && Ptime.equal ua1 ua2
+      && eq_key
+           (X509.Certificate.fingerprint `SHA256 cer1)
+           (X509.Certificate.fingerprint `SHA256 cer2)
+      && eq_key
+           (X509.Private_key.encode_der pk1)
+           (X509.Private_key.encode_der pk2)
+  | _ -> false
```
hannesm: see below, also will need to be implemented. but it is a good default :)

tests/storage/tests.ml
PizieDust: I just added a commit to compare lists with `List.equal`. 
https://github.com/robur-coop/mollymawk/pull/275/commits/23c9f36800c0c19676abe95fce004e148d95035c

I think this can solve the problem

PizieDust (commented):
@hannesm thank you. I will go ahead to merge this in.

PR #276, opened 2026-08-22
refactor storage and our block functor

PizieDust:
This PR decouples and isolates IO functionality from the data operations.

### DIFF
--- a/src/storage.ml
+++ b/src/storage.ml
@@ -7,7 +7,15 @@ let current_version = 10
    10 we now have scaling policies for unikernels, default is no policy for existing unikernels
 *)
 
-type t = User_model.user list * Configuration.t list * Utils.Email.t option
+type t = {
+  mutable users : User_model.user list;
+  mutable configurations : Configuration.t list;
+  mutable email : Utils.Email.t option;
+}
+
+let configurations { configurations; _ } = configurations
+let email { email; _ } = email
+let users { users; _ } = users
 
 let t_to_json ?(version = current_version) users configurations email =
   `Assoc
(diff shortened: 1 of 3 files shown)


src/storage.ml
```diff
@@ -154,3 +162,73 @@ let count_active users =
 
 let count_superusers users =
   List.length (List.filter (fun u -> u.User_model.super_user) users)
+
+let configruation_label_eq (c1 : Configuration.t) (c2 : Configuration.t) =
```
hannesm: typo `config_ur_ation`, not `config_ru_ation`.

src/storage.ml
```diff
@@ -154,3 +162,73 @@ let count_active users =
...
+  Vmm_core.Name.Label.equal c1.name c2.name
+
+let exists (configurations : Configuration.t list)
+    (configuration : Configuration.t) =
+  List.exists (configruation_label_eq configuration) configurations
+
+let store_email t email =
+  t.email <- email;
+  t
```
hannesm: if you assign, why return the t that came in?

src/storage.ml
hannesm: I'd name it `configuration_name_eq`.

src/storage.ml
```diff
@@ -154,3 +162,73 @@ let count_active users =
...
+
+let insert_configuration t (configuration : Configuration.t) =
+  if exists t.configurations configuration then
+    Error
+      (Fmt.str "configuration %s already exists"
+         (Configuration.name_to_str configuration.name))
+  else (
+    t.configurations <- t.configurations @ [ configuration ];
+    Ok t)
```
hannesm: same - why a `t`?

I usually follow the API of:
```OCaml
val whatever : t -> x -> t
```

if the result is a fresh `t`. Taking an argument and returning it is not very useful -- I already have it at the call site.

unikernel.ml
```diff
@@ -293,13 +293,14 @@ struct
     in
     go (Map.empty, []) m
 
-  let generate_csrf_token store user now reqd =
+  let generate_csrf_token store storage_data user now reqd =
```
hannesm: now that seems very unpleasant to have now "store" and "storage_data". Aren't there other solutions?

PizieDust: maybe we keep it as `store` which has the type  (`Stored_data.t * Storage.t`) and then do ` (fst store)` and when we need the data we do` (snd store)`?

unikernel.ml
```diff
@@ -3372,7 +3393,7 @@ struct
   let unikernels_stats stack instance store =
     let cb name st =
       Lwt.async (fun () ->
-          handle_stats stack instance st name store >>= function
+          handle_stats stack instance st store name >>= function
```
hannesm: I'd avoid changing the argument order here...

unikernel.ml
```diff
@@ -3274,14 +3295,14 @@ struct
             prune_clone stack albatross group ~unikernel_name ~clone_to_kill
               ~user_name)
 
-  let handle_stats stack state ((rusage, _, _) : Vmm_core.Stats.t) name store =
+  let handle_stats stack state ((rusage, _, _) : Vmm_core.Stats.t) store name =
```
hannesm: see below, no reason for this change in this PR as far as I can tell.

hannesm:
apart from the `handle_stats` argument ordering, good to go. thanks!

PizieDust:
> apart from the `handle_stats` argument ordering, good to go. thanks!

Thank you for the review and the ideas :) @hannesm

PR #278, opened 2026-09-15
Addresses review comments from https://github.com/robur-coop/mollymawk/pull/277

PizieDust:
This PR addresses: https://github.com/robur-coop/mollymawk/pull/277#discussion_r3988075113 and https://github.com/robur-coop/mollymawk/pull/277#discussion_r3988065221

For the raw dump, I debloated it by removing for instance: if a user ha 10 cookies, I remove 8 and leave 2.


hannesm:
thanks

PR #279, opened 2026-09-15
Tests for authentication functions

PizieDust:
This PR adds tests for some functions we use during authentication:

- `User_model.create_user`
- `Storage.find_by_name`
- `Storage.find_by_email`
- `Utils.Email.validate_email`
- `User_model.password_validation`
- `User_model.login_user`
- `User_model.update_user`
- `User_model.verify_email_token`
- `User_model.is_email_verified`


hannesm:
I think the tests are fine. I'm not sure they're testing interesting things (check_valid_registration looks slightly boring)... but I guess that is fine.

I am a bit surprised about the structure of the tests, the amount of stuff in the "dune" file is surprising. why does autoscalar.ml is copied?

PizieDust:
> I think the tests are fine. I'm not sure they're testing interesting things (check_valid_registration looks slightly boring)... but I guess that is fine.
> 
> I am a bit surprised about the structure of the tests, the amount of stuff in the "dune" file is surprising. why does autoscalar.ml is copied?

autoscaler is used in a test I haven't pushed yet. it should have been included in the relevant PR.

hannesm:
since this is better than the state before, let's merge this.

PR #281, opened 2026-09-17
Test mock devices and api endpoints

PizieDust:
This Pr tries to instantiate mollymawk's functor with some fake devices and runs in memory. 
It incldues more tests for the registration endpoint. 
The interesting function here is `Mock_devices.query_endpoint` which tries to feed raw HTTP requests into the H1 server and capture responses in memory.


tests/api/dune
```diff
@@ -1,5 +1,8 @@
 (test
  (name api_tests)
+ (enabled_if
+  (= %{context_name} "default"))
+ (flags :standard -warn-error -a)
```
hannesm, line 5: why?

tests/api/dune
```diff
@@ -12,14 +15,29 @@
...
+  colombe.emile
+  dns-client-mirage
+  happy-eyeballs-mirage
+  tls-mirage
+  mirage-runtime
+  multipart_form
+  multipart_form-lwt))
 
-(copy_files# ../../autoscaler.ml)
+(copy_files# ../../*.ml)
```
hannesm, line 43: why? why not being explicit here, but copying everything?

hannesm:
so what is the general approach? should these tests be working in a different unikernel which doesn't use mirage anymore? which part of the API is tested and at which level?

tests/utils/test_utils.ml
```diff
@@ -450,3 +450,25 @@ let make_mock_user ?(name = "testuser") ?(email = "test@example.com")
...
+
+let string_contains ~sub str =
+  let len_sub = String.length sub in
+  let len_str = String.length str in
+  let rec check i =
+    if i + len_sub > len_str then false
+    else if String.sub str i len_sub = sub then true
+    else check (i + 1)
+  in
+  check 0
```
hannesm: is this String.includes? (since 5.5)

tests/api/dune
PizieDust: unikernel.ml calls many modules in the different files. listing each one explicitly was a lot of files so I just used the wildcard

tests/api/dune
PizieDust: when we copy the files and try to run the test wihtout disabling, the tests don't run and dune short-ciruits to display the errors

```
File "autoscaler.ml", line 115, characters 20-23:
Error (warning 27 [unused-var-strict]): unused variable err.
File "console_viewer.ml", line 3, characters 26-39:
Error (warning 27 [unused-var-strict]): unused variable instance_name.
File "albatross.ml", line 397, characters 12-15:
Error (warning 27 [unused-var-strict]): unused variable hdr.
File "monitoring.ml", line 445, characters 32-33:
Error (warning 27 [unused-var-strict]): unused variable l.
File "unikernel_scaling.ml", line 49, characters 34-35:
Error (warning 27 [unused-var-strict]): unused variable p.
```

PizieDust: we can remove the flag later when we handle these minor warnings of unused variables

tests/api/dune
PizieDust: 40 fiiles

tests/utils/test_utils.ml
PizieDust: Thanks. i had to upgrade my switch.

PizieDust:
> so what is the general approach? should these tests be working in a different unikernel which doesn't use mirage anymore? which part of the API is tested and at which level?

the idea is to have fast tests for mollymawks api's. since Unikernel.Main uses mirage signatures (Mirage_block.S, Mirage_kv.RO, and Tcpip.Stack.V4V6), we instance it with lightweight in-memory mock implementations  and feed raw HTTP requests directly into the H1.Server_connection and drain the response iovecs into a buffer. 

Parts of the API tested so far is only the registration endpoints. (in this pr). 
to keep the pr small, it's better to have other pr's test the other endpoints.

what we really tests is:
- session cookie extraction (molly_session), CSRF validation (molly_csrf), and soon admin permission checks.                                                                                                                 
- enforcing system guards (e.g. forbidding the removal of the last administrator or deactivating the last active user).                                                                                                                                                                                                                                         
- verifying HTTP status codes (200 OK, 400 Bad Request, 403 Forbidden, 404 Not Found) and json responses.

later when we refactor, it can easy to know if we broke some functionality

hannesm:
so should these tests be portable when mollymawk is refactored?

what about albatross interaction?

hannesm:
thanks, this looks fine to me. feel free to merge this.

